PR #82, opened 2020-07-09
Documentation Daniel

drg101:
I can't run the command to generate documentation because ./node_modules isn't backed up by git. Is the documentation.js a module or is placed there manually?


kevin-bruhwiler:
I think you have to install it. "npm install documentation"

kevin-bruhwiler:
I think you have to install it. "npm install documentation"

kevin-bruhwiler:
You also have to update the wiki manually with the new docs.

PR #109, opened 2020-10-12
Dynamic queries

kevin-bruhwiler:
@drg101 can you take a look at this? I made quite a lot of changes, and some of it is a bit hacky, but I'm not sure how to generalize things until we get the metadata catalogue working on the client-side.

Let me know what you think, or if anything needs to be changed.


drg101:
@kevin-bruhwiler Is there a good place/settings to test the heat waves? Also, the hacky stuff is fine, we should rethink the whole menu system once we get the new query infrastructure and metadata catalog.

kevin-bruhwiler:
Don't really know. If you put the sliders down you'll get some data for fort collins. Generally you see more heat waves in the south.

drg101:
I see, is it supposed to query just when you change settings? It seems like it only runs queries when you do that, not when you are panning around like the rest of the features.

Also, I think for the same reason as above, you can run queries without the checkbox being selected.

Lastly is what does this data represent? I wonder if you could customize the popup to have relevant info.

kevin-bruhwiler:
I could customize the pop-up, although it will just have the same info as the sliders.

It was updating on move, but apparently I broke that fixing something else, and after playing around with it this morning I think I've figured it out. When the checkbox is checked or unchecked the "updateMaps" event is called (which I feel like it shouldn't be) but it gives the wrong value for whether a box is checked. For example, if you uncheck the box, that event will re-run the query while claiming the box is still checked **after** the data has been cleared from the box being unchecked. I'm reading the "checked" value directly from the dom, so I guess the "updateMaps" event is firing before the "checked" value is updated, but somehow after the event for it being unchecked is fired? I really don't understand what's causing the "updateMaps" event to be fired though. When the map is moved it fires with the correct value, that part I understand.

It also looks like some of the tests have the number of datasets that are displayed hard-coded in? I think those may need to be removed, since we'll be adding lots of datasets.

kevin-bruhwiler:
Actually, it looks like the "updateMaps" event is being fired any time any part of the map is clicked, even white space on the menu. Is that intentional?

kevin-bruhwiler:
@drg101 @tlujan94 Can you both take a look at this when you have time? Let me know if you spot any obvious problems, and let me know if you have any questions about how the new queries work. We'll be moving all the queries the new format at some point.

drg101 (approved):
The only thing I notice is that if you press enter on the year selector it reloads the iframe for some reason for me. I think its a live-reload thing though.

PR #110, opened 2020-10-14
Added NoUi Sliders

drg101:


drg101:
@kevin-bruhwiler Could you check out this implementation? I basically moved everything to the JSON and used a UI library. I changed your code to access the new sliders as a test and I believe it's working perfectly.

kevin-bruhwiler:
Bearing in mind that I don't know exactly what the data should look like, it does seem to be functioning as expected.

kevin-bruhwiler:
I think a lot of these tests that depend on the json file having certain values need to be removed. We'll be editing that file a lot.

drg101:
I switched the tests to a test-only JSON, I also changed added a second argument to updateFutureHeat, which specifies if constraints have been changed. If that argument is false, the map won't be cleared. 

The JSON that describes the constraints has an attribute specifically for what function to call when a constraint has been changed now.

This is just a work-around, but it works for now and should make the queries look much cleaner.

kevin-bruhwiler:
The smarter re-querying is nice, and I don't see any immediate issues with specifying the function that has to be called. I don't know if it's possible to write a single function that can handle every use case.

Only issue I see is that it seems like you have to toggle the heat wave checkbox a couple of times before it will start querying data.

drg101:
Just fixed that issue, thanks for pointing that out.

kevin-bruhwiler (approved):

PR #113, opened 2020-11-12
added methods for creating and executing compound queries with more t…

kevin-bruhwiler:
…han one query

### DIFF
--- a/src/iframe/js/grpc/GRPC_Querier/grpc_querier.bundle.js
+++ b/src/iframe/js/grpc/GRPC_Querier/grpc_querier.bundle.js
@@ -153,7 +153,7 @@ function fromByteArray (uint8) {
 }
 
 },{}],2:[function(require,module,exports){
-(function (Buffer){
+(function (Buffer){(function (){
 /*!
  * The buffer module from node.js, for the browser.
  *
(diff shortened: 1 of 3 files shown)


kevin-bruhwiler:
Can you both take a look at this? There aren't many changes, but I want to make sure that it makes sense to you, and that this framework will work for the new sorts of queries you're putting together.

I can't properly test this until Menuka resolves some inconsistencies with GISJOIN fields in the database (which he's been promising to do since Monday) so I can't verify that it's totally working.

src/iframe/js/grpc/GRPC_Querier/grpc_querier.bundle.js
```diff
@@ -2674,6 +2646,82 @@ SustainQuerier = {
...
+     * @param {Object} secondQuery
+     *        The second query object, if present
+     * @param {Object} secondCompoundRequest
+     *        The second compound request object, if present
+     * @param {Number} join
+     *        The type of join being used (number corresponding the proto definition)
+     * @return {Object}
+     *         The compound request object
+     */
+    makeCompoundQuery: function (firstQuery, firstCompoundRequest, secondQuery, secondCompoundRequest, join=0){
```
drg101, line 2694: So, in essence, you could create an infinite chain of queries using this function right?

kevin-bruhwiler: That's the idea. Let's you query as many collections as you need to, all at once. It's essentially a binary tree, with each node specifying how to join the leaves below it.

drg101: Seems simple enough, this should make creating new queries with the UI easy too.

kevin-bruhwiler:
Ok, I think this is ready for review now. I've verified that it works and I've added an example query. It's a little cleaner and hopefully easier to write. If you run into any bugs with the new server let me know.

drg101 (approved):
Looks clean and seems faster than before, good work!

PR #114, opened 2020-11-18
New interface

drg101:


drg101:
This uses my covid query from the video, I'm not exactly sure what keegans does still

drg101:
And the SVI is the flawed one, not sure if I should put "Estimated SVI" or something

kevin-bruhwiler (approved):
I think calling it SVI is fine for now.

PR #122, opened 2021-01-08
Auto Generate Menu

drg101:


drg101:
@honeybadgerofdoom @Marc-9 @pierce-smith1 Wait until after tonight's meeting to start reviewing

src/iframe/js/library/apertureUtil.js
```diff
@@ -0,0 +1,340 @@
...
+    },
+    /**
+     * What type is a GeoJSON feature?
+     * @memberof Util
+     * @method getFeatureType
+     * @param {object} feature GeoJSON feature, a latlng point will be extracted. Can be a point, linestring, or polygon.
+     * @returns {number} Enum from FEATURETYPE or -1 if not found
+     */
+    getFeatureType: function (feature) {
+        if ((feature.geometry) && (feature.geometry.type !== undefined) && (feature.geometry.type === "Polygon")) {
```
Marc-9, line 49: Unless you are not a fan of nested if statements could be a bit cleaner to do

if((feature.geometry) && (feature.geometry.type !== undefined)){

// Here your 3 ifs (feature.geometry.type === "Polygon") etc etc
//else ret -1

}
else{
return -1
}

drg101: Fixed that, thanks for pointing it out. That whole library was written last summer and definitely needs a freshening up.

PR #123, opened 2021-01-16
Small tweaks

Marc-9:
Just a couple of small quality of life fixes, not all of them need to be changed though, and there are still some more things I would like to change.
So far I have altered the css for the menu so that the sliders don't go on top the the header, as well as alter the width and added some js to close the menu when the map is clicked on, but since queries are made every time the menu is closed this many not be wanted

### DIFF
--- a/src/iframe/css/third-party/leaflet-sidebar.css
+++ b/src/iframe/css/third-party/leaflet-sidebar.css
@@ -2,7 +2,7 @@
   position: absolute;
   top: 0;
   bottom: 0;
-  width: 100%;
+  width: 55em;
   overflow: hidden;
   z-index: 2000;
 }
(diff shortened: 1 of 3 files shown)


Marc-9:
When clicking on an icon or tract leaflet map no longer flies to it and alters bounds

drg101:
Made the flyTo animation just go to the center of where the user clicks. This fixes the bug for the icons (on my end), whilst removing the zooming which I think was what was annoying about the old flyTo.

drg101:
I think this allows the map to feel as "responsive" as possible whilst also being super simple.

drg101 (approved):
Everything seems to be an improvement, looks good.

PR #125, opened 2021-01-17
Add Pipes

drg101:
Added pipe.js and documentation/tests for it.

### DIFF
--- a/src/dependencies/pipe.js
+++ b/src/dependencies/pipe.js
@@ -65,7 +65,7 @@ Pipe = {
     /**
       * Finds the index of a pipe
       * @memberof Pipe
-      * @method pipe
+      * @method indexOfPipe
       * @param {string} id name/id of pipe to send data over
       * @returns {Number} index of pipe if found (>= 0). -1 if not found.
       */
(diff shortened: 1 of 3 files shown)


Marc-9 (approved):
Looks good to me, I would like to see some example code of how this would look in action

src/dependencies/pipe.js
```diff
@@ -0,0 +1,85 @@
...
+            return true;
+        }
+        return false;
+    },
+
+    //private functions --------------
+    /**
+      * Finds the index of a pipe
+      * @memberof Pipe
+      * @method pipe
```
pierce-smith1, line 68: The method name in the docs here is wrong.
Other than that I don't see a problem. Tests pass, code is clean.

drg101:
Made the flyTo animation just go to the center of where the user clicks. This fixes the bug for the icons (on my end), whilst removing the zooming which I think was what was annoying about the old flyTo.

drg101:
I think this allows the map to feel as "responsive" as possible whilst also being super simple.

PR #127, opened 2021-01-19
Render full-screen map button

mehrotrasan16:
Map2 already had the fullscreenControl option, but it wasn't being used. 
Modified map2.html and package.json: added leaflet-fullscreen package to show a fullscreen button below the zoom buttons.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "jsdom-global": "^3.0.2",
     "leaflet-timedimension": "^1.1.1",
     "leaflet.markercluster": "^1.4.1",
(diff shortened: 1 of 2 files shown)


drg101:
The menu button disappears when you click fullscreen. I don't really see the advantage of this over just clicking F11, as it does the same things.

mehrotrasan16:
> The menu button disappears when you click fullscreen. I don't really see the advantage of this over just clicking F11, as it does the same things.

Oh that's weird.  Guess I didn't see that breaking. My bad, I'll close this pull request, until I figure out a way to fix that.

PR #129, opened 2021-01-23
Graph overlay

Marc-9:
Ya I know we wanted Iframes, however this actually has way more advantages

It can't take up as much space as the user wants, as they can resize it and move it on the screen wherever they please
Now we don't need to pipe infrastructure data into the other iframe reducing duplication costs

Bug when user tries to move the overlay to fast and highlights the page


drg101:
I definitely like this functionality, as it will be super useful for all sorts of widgets, but I do foresee a few issues for graphing:

- The graphing functionality displays what is visible on the map. This widget would need to cover a decent portion of the map to be useful for displaying graphs, meaning a good portion of what is being graphed wouldn't be visible.
- The widget would need to big enough for graphs, but not cover the whole map, meaning that there would need to be heavy resizing restrictions.
- This is more of a personal thing, but it is a bit awkward for a part of the map to be covered in general (the way the menu is right now kind of annoys me for this reason), I've never seen it on other Geospatial tools.

drg101:
But of course, I have no idea how we could make something better than this. I would hate for the map to be half the size it currently is.

pierce-smith1:
I think some of these issues would be less egregious with a (convenient) way to collapse the graph. Perhaps something on the bottom of the popup that can be clicked to "fold" the graph into a single, much smaller element that can be expanded again, kind of like the blinds feature in GNOME. That way, if a user is annoyed by how they can't see the graph, they can just push it out of the way until they want it again.

Marc-9:
In response to Daniel and Pierce
Yes atm it looks super awkward to have a box forever taking up space, the plan is to have it hidden by default with an option to expand and collapse it similar to the menu.
Also the background will be more translucent
I think a resizable overlay has many advantages than an Iframe as mentioned above
Then there should be some semi-easy way to auto realize the overlay to fit the content within the div

Marc-9:
Still not complete just updating the progress

PR #134, opened 2021-01-27
Map data filter

### LINKED ISSUE #132: System to Filter & Arrange Data for Graphs

pierce-smith1:
Adds a new class, `MapDataFilter`, which data can be passed through to prepare it for graphing.
`graph.js` contains an example of how it can be used. 

Closes #132.


drg101 (approved):
Look really good, I think the optimization functions will come in handy for some other stuff too.

PR #143, opened 2021-02-10
Fix graph.png Source

drg101:


src/iframe/map2.html
```diff
@@ -68,7 +68,7 @@
 
 
             <div class="btn-group" role="group" aria-label="Basic example">
-              <button type="button" id='nav-graph-button' class="btn btn-outline-dark" disabled><img src="/images/graph.png" class="graph-button-pic"></button>
+              <button type="button" id='nav-graph-button' class="btn btn-outline-dark" disabled><img src="../../images/graph.png" class="graph-button-pic"></button>
```
drg101, line 71: I had to change this to the relative path, as the absolute path depends on the environment the site is deployed on.

PR #147, opened 2021-02-16
Seriously actually restore Matt's ui

pierce-smith1:
Fix a problem that caused Matt's UI to disappear from dev... for real this time.


Marc-9 (approved):
Really actually working now I promise

PR #150, opened 2021-02-19
Update dev with multiple chart progress

pierce-smith1:
This obviously isn't _done_ yet, but we need it in dev.

### DIFF
--- a/src/iframe/css/IframeStyle.css
+++ b/src/iframe/css/IframeStyle.css
@@ -401,7 +401,3 @@ body {
   flex-direction: column;
 }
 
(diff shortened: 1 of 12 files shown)


Marc-9 (changes requested):
This can't be merged because I want mine merged first sou handle the merge conflicts

Marc-9 (changes requested):
You made min chart size as 300 in chart area.js but the min size of the resizable was not changed in resizable.js

pierce-smith1:
Temporarily closing this because I'm working on scatterplots.

PR #151, opened 2021-02-19
Functional graph button in menu

Marc-9:
The overlay graph now works with the button Matt has implemented
This PR also removed unused libraries such as, leaflet easy button (the star buttons that controlled the graph) and leaflet sidebar since we are using Matts sidebar

This 100% should be merged before Pierce's so that he has to deal with all the merge conflicts that will follow


Marc-9:
Oops meant to go into Dev not master

PR #157, opened 2021-02-25
Fix county querying

pierce-smith1:
Resolves a bug where county GISJOIN information was not being properly queried (i.e. not properly implementing hacky workaround to deal with mongodb nonsense).


Marc-9 (approved):
Quite the interesting bug,  however it appears to be resolved with this branch

drg101 (approved):
Works

PR #158, opened 2021-02-25
Added Skew Correcting on Color Coding

drg101:
Made it so you can correct for skewing in a dataset during color-coding.

Here is the basic idea: https://www.desmos.com/calculator/px5vxgzujw
In that example, s = how skewed the value you are color coding on is.
Green = correction for right skewness.
Purple = correction for left skewness.
Red = how it is if you don't set skew for a dataset or skew is set to 0

### DIFF
--- a/src/iframe/js/library/autoQuery.js
+++ b/src/iframe/js/library/autoQuery.js
@@ -379,20 +379,20 @@ class AutoQuery {
       * @returns {string} hex color code
       */
     getColor(properties) {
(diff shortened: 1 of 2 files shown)


drg101:
closes #155

drg101:
Going to update to this version: https://www.desmos.com/calculator/px5vxgzujw

Marc-9 (approved):
idk looks the same to me

PR #160, opened 2021-02-25
Administrative Tracts may now overlap

### LINKED ISSUE #159: Allow Multiple Layers of Same Resolution at Once

drg101:
Made it so different county/tract features don't get rejected by `renderGeoJson` by setting the id to a combination of the data, and the county/tract id.

closes #159

### DIFF
--- a/src/iframe/js/library/autoQuery.js
+++ b/src/iframe/js/library/autoQuery.js
@@ -233,11 +233,12 @@ class AutoQuery {
                 return;
             Util.normalizeFeatureID(GeoJSON)
             GeoJSON.id = `${GeoJSON.properties}_${data.id}`
+            if (this.layerIDs.includes(GeoJSON.id))
+                return;
             GeoJSON.properties = {
                 ...GeoJSON.properties,
                 ...data
             }
-            console.log(GeoJSON.properties)
             data = GeoJSON;
         }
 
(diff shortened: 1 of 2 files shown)


drg101:
Noticed bug

src/iframe/js/library/autoQuery.js
```diff
@@ -228,10 +228,13 @@ class AutoQuery {
       */
     renderData(data, forcedGeometry) {
         if (this.linked) {
-            const GeoJSON = this.backgroundLoader.getGeometryFromGISJOIN(data.GISJOIN, forcedGeometry);
+            const GeoJSON = JSON.parse(JSON.stringify(this.backgroundLoader.getGeometryFromGISJOIN(data.GISJOIN, forcedGeometry)));
```
pierce-smith1, line 231: Why?

pierce-smith1 (approved):
I have no idea how any of the code you wrote makes this work, but it does in fact work.
Edit: Now that I've bothered to read the PR description it makes more sense.

src/iframe/js/library/autoQuery.js
drg101: I had to deep clone the data I got back because I was changing the `id` of the reference, which ended up causing problems.

PR #167, opened 2021-03-02
Self hosted the react code

drg101:


Marc-9 (approved):
Every day we stray further from god, the amount of data we are forcing these poor browsers to load is so sad

PR #171, opened 2021-03-08
Fast Query System

drg101:
Changed a lot of things. Tracts are very, very fast now. Counties need to be simplified on the server-side, but when they are, we will have crazy fast queries.

switched geometryLoader to be a worker that queries based on geohashes. 

@pierce-smith1 I had to change some queryWorker code to get things working


src/iframe/js/library/GeometryLoaderJob.js
```diff
@@ -0,0 +1,44 @@
+/**
```
drg101, line 1: btw, this file isn't used, but I will use it in a change very soon.

src/iframe/map2.html
```diff
@@ -118,6 +117,7 @@
     <script src="js/ui/modeling/components/ModelResolution.js"></script>
 
     <!-- Frame JS, Load Last! -->
+    <script src="js/frame-specific/cool-guy.js"></script>
```
Marc-9: this file does not exist

src/iframe/js/library/smartQuerier.js
```diff
@@ -212,6 +212,7 @@ class SmartQuerier {
       * @method addToCache
       */
     addToCache(collection, data) {
+        return; //disabling this for until we can standardize things between query libs
```
Marc-9, line 215: just comment out the code so we don't have errors in console

src/iframe/js/library/smartQuerier.js
```diff
@@ -133,6 +132,7 @@ class SmartQuerier {
       */
     wrapResponseCallback(collection, callback) {
         let onResponseCallback = (response) => {
+            console.log("hereee")
```
Marc-9, line 135: remove

src/iframe/js/library/boundsToGISJOIN.js
```diff
@@ -0,0 +1,98 @@
...
+    buckets: null, //loaded at config
+    geohashResolution: 3,
+
+    config: function (collection) {
+        const jURL = collection === "tract_geo_140mb" ? pathsToBuckets.tracts : pathsToBuckets.counties;
+        fetch(jURL)
+            .then(response => response.json())
+            .then(data => {
+                BoundsToGISJOIN.buckets = data
+                console.log(`Loaded ${collection} buckets`)
```
Marc-9, line 16: Remove

Marc-9 (changes requested):
There also is a console.log in query worker that needs to be removed
Honestly, the site was not faster for me at all could not even load all the counties in colorado
Also I assumed the grapher could graph any data from any constraint. maybe the origin of this pull is form an old branch, but also the grapher could not handle it when I tried to graph lots of future heatwave data

Marc-9:
Also I need a quick run down on the graph querier code

drg101:
@Marc-9 Update the branch, I resolved all of your issues. 
Try the county query now, I switched the file from the ~200mb to the ~30mb one. This PR won't fix all the issues with the query system, just all the client-facing ones.

drg101:
I also don't really understand much of the graph code, but I think some optimizations will definitely be needed with this new system.

drg101:
> Honestly, the site was not faster for me at all could not even load all the counties in colorado

@Marc-9 Make sure you are not using the heatwaves data as a benchmark, as it doesn't include much of Colorado. Use the risk index instead.

Marc-9 (approved):
All my issues and comments were addressed

src/iframe/js/library/autoQuery.js
```diff
@@ -7,6 +7,7 @@
  */
 
 class AutoQuery {
+    static queryWorker = new SharedWorker('js/library/queryWorker.js', {name: "Auto query worker"}); //init querier
```
pierce-smith1, line 10: Out of curiosity, why is this static now? Were things breaking if multiple AutoQuery objects were made?

pierce-smith1 (approved):
This is insane

honeybadgerofdoom (approved):
Ludicrous speed indeed!

src/iframe/js/library/autoQuery.js
drg101: There was just no reason to make more than one, as I am setting sessionID's with each request anyways.

PR #172, opened 2021-03-09
Use IndexedDB & Add Zoom Restriction Option

drg101:
Use IndexedDB, with the Dexie wrapper, to cache shapefiles. Advantages being:
- Doesn't use JS heap at all.
- The cached data will stay on the User's machine through reloads.

Also added ability to restrict zoom levels. Look at static attributes of `AutoQuery`. I've added a `DEV` boolean to map2.js where we set those zoom things to a less restricted amount.


drg101:
Gonna add the aggregation thingy

src/iframe/js/frame-specific/map2.js
```diff
@@ -1,8 +1,11 @@
 const MAPNUMBER = 2;
 const e = React.createElement;
 
-//const queryAlertText = document.getElementById('queryInfoText');
-
+const DEV = true;
```
Marc-9, line 4: Wouldn't it be better to parse the domain url and check
Something like
const DEV = window.location.href.split("/")[1] === "dev"

Marc-9 (changes requested):
Errors with Auto generated constraints
Electrical Substations- Uncaught TypeError: feature.properties is undefined
Power Plants/Hospitals- Uncaught Error: Invalid GeoJSON object.
Gas Pipelines- Uncaught TypeError: feature.properties is undefin in aperture util
Pretty much all the other also spam the console with invalid geojson object

src/iframe/js/frame-specific/map2.js
drg101: Yea, good idea

drg101:
@Marc-9 addressed all of your issues

drg101:
Going to add preloading

drg101:
This thing is crazy fast now.

Marc-9 (approved):
Graph does not work, perhaps this will be fixed when merged
Also small css issue with the menu on screens <= 13 inches

pierce-smith1:
Is zooming out supposed to be blocked in DEV mode? Because it doesn't seem to be.
Also, where is the preload status text showing up (is it supposed to show up at all?)

pierce-smith1:
No idea why charting doesn't work because the code that deals with it hasn't changed, but we're going to have to deal with this problem regardless once the complete charting feature gets merged into dev. Boy, that's going to be fun.

drg101:
@pierce-smith1 
> Is zooming out supposed to be blocked in DEV mode? Because it doesn't seem to be.
> Also, where is the preload status text showing up (is it supposed to show up at all?)
Zooming is unrestricted in DEV mode. The queries are honestly fast enough to zoom out a ton now anyways. 



> No idea why charting doesn't work because the code that deals with it hasn't changed, but we're going to have to deal with this problem regardless once the complete charting feature gets merged into dev. Boy, that's going to be fun.

All the data still goes through renderInfrastructure's renderGeoJson method, so that's weird. Which attributes are you pulling from the data, I've filtered the response data to only include what is going to be on the popup, which is likely your issue.

drg101:
> Graph does not work, perhaps this will be fixed when merged
> Also small css issue with the menu on screens <= 13 inches

@honeybadgerofdoom Could you address the skinny menu issue in a separate PR?

PR #177, opened 2021-03-11
Added persistence prompt

drg101:


Marc-9:
When we decide on a time to implement this feature, testing in Firefox will be simple
TO test on chrome we will need a console log to see if we passed chromes odd permission request specifications, could cause some issues as testing on localhost will provide separate results as testing on the urban-sustain website
Additionally new users to the site may have the request fail as compared to us who visit the site constantly

drg101:
Will push at a later time.

PR #178, opened 2021-03-12
Switch from SharedWorker to Worker

drg101:


src/iframe/js/frame-specific/map2.js
```diff
@@ -274,7 +272,7 @@ parent.setterFunctions.push({
     setterFunc: thisMapsSetter,
     mapNum: MAPNUMBER
 });
-//let chartSystem = new ChartSystem(map, "json/graphPriority.json");
+let chartSystem = new ChartSystem(map, "json/graphPriority.json");
```
pierce-smith1, line 275: Any particular reason why you uncommented this? (why was it even commented out in the first place?)

pierce-smith1 (approved):
All kinds of queries still work, so it seems nothing is broken

src/iframe/js/frame-specific/map2.js
drg101: By accident

PR #185, opened 2021-03-15
Made Loading Prettier

drg101:
You can test the loading by deleting the indexedDB's from your browser. In Chrome, it's under `Application`.


pierce-smith1:
This is _really_ cool. I'm not sure about the dark blue color for the loading bar, though, I think a slightly lighter and less saturated blue would look a lot nicer.

drg101:
> This is _really_ cool. I'm not sure about the dark blue color for the loading bar, though, I think a slightly lighter and less saturated blue would look a lot nicer.

@pierce-smith1 Will change, thanks for pointing that out.

pierce-smith1 (approved):
Looks much nicer

PR #186, opened 2021-03-15
Multiple Charting Prototype

pierce-smith1:
This updates dev with our current progress for multiple charting. The interface is NOT refined yet, but it should at least work.

This PR needs to be made so that there is some amount of functional charting in dev for the next demo.

A couple of features will not be chartable because the charting catalog has not updated, e.g. risk index. Future heatwaves, county SVI, and median tract age are a couple of properties that SHOULD still work, so test those.

Line charts do not work yet. They will eventually be used for COVID sliding window data.


drg101 (approved):
It doesn't break any existing functionality, but here are a few things I noticed that should be addressed asap after merging:
- After removing a layer, like future heatwaves, it seems that the data stays in the JS heap.
- The site feels a bit slower, so we should put the filtering into a worker.
- This error is very annoying: Error: <rect> attribute x: Expected length, "NaN".

PR #190, opened 2021-03-16
Charting QOL

pierce-smith1:
This PR fixes some (but not all) QOL problems with charting.

- Scatterplot axes have human-readable names
- A message informs the user when there aren't enough constraints to graph
- Spacing on graph titles and labels is improved
- No more "got NaN for attribute" errors

I wanted to get plotting to work for risk index, but since that data should be presented in a qualitative instead of quantitative fashion I need a completely new graph type which probably won't be ready in an hour.


src/iframe/js/library/charting/scatterplotManager.js
```diff
@@ -112,8 +120,28 @@ class ScatterplotManager {
...
+        }).constraints[this.currentFeatures.x].label;
+
+        let readableYName = this.catalog.find(e => {
+            for (let constraint in e.constraints) {
+                if (constraint === this.currentFeatures.y) {
+                    return true;
+                }
+            }
+            return false;
+        }).constraints[this.currentFeatures.y].label;
```
drg101, line 140: What happens if label is undefined?

drg101 (approved):
Getting this error in chrome now: `d3.min.js:2 Error: <svg> attribute viewBox: A negative value is not valid. ("0,0,-61,317")`

This is better than it was, just keep up the incremental fixes and it should be good soon.

src/iframe/js/library/charting/scatterplotManager.js
pierce-smith1: The scatterplot axis has no name

Though if this does happen, that means the charting catalog is ill-formed

pierce-smith1:
> Getting this error in chrome now: `d3.min.js:2 Error: <svg> attribute viewBox: A negative value is not valid. ("0,0,-61,317")`

Meanwhile, completely silent in firefox :/

PR #193, opened 2021-03-17
Filter Worker

drg101:
@pierce-smith1 I had to edit some code in `Util` to make this work, turns out that when you pass an object to a worker all the `proto` stuff dies. I also couldn't import leaflet on the worker cause there's no DOM (which was quite the rabbit hole).

Anyways, it works and it stutters a bit less for me.


pierce-smith1 (approved):
Doesn't seem to have broken anything and the site definitely runs a bit faster. No issues with the code changes - it's interesting that getValues() is now async but I don't envision that being an issue at all.

PR #196, opened 2021-03-18
Random Lattice Num

drg101:
Added a `DEV` log which will tell you which machine you are connected to. I had to pass the number to the workers on config.


drg101:
I added a backdoor to target specific servers. You can use it by setting `latticeNum` in localStorage as the lattice machine you want.

pierce-smith1:
`SmartQuerier` is still hard-coded to `lattice-46` and port `27017`, is this an entirely different database?

pierce-smith1 (approved):
Approved on the condition that `lattice-46` still being hard-coded in SmartQuerier is intentional

drg101:
Those statics don't get sent with the query anymore so they don't affect anything luckily.

PR #200, opened 2021-03-19
Make Other Clustering Work

drg101:
Just a fix


menuka94 (approved):

Marc-9 (approved):
Runs fine, but graphing team needs to fix
Uncaught TypeError: xfeat is undefined
In scatterplot.js

PR #201, opened 2021-03-20
Dark mode switch bug fix

honeybadgerofdoom:
This dark mode switch no longer stacks on top of the nav bar on small screens because I actually took the time to put everything into a grip system.


honeybadgerofdoom:
I also moved the JS that handles to nav buttons into their own file to clean up map2.js a little bit

pierce-smith1 (approved):
Works:tm:

drg101 (approved):
Good

PR #204, opened 2021-03-21
Fix "xfeat is undefined" bug in scatterplots

pierce-smith1:
Remember when I said I didn't think getValues() being async would cause any problems?


drg101 (approved):
Good

PR #205, opened 2021-03-21
Proximity Queries

drg101:
Just added a dataset, its actually pretty cool, thanks @menuka94


menuka94 (approved):

pierce-smith1 (approved):
I like this, should be easy to put into the charting catalog too since it's all numeric.

PR #233, opened 2021-03-21
Charting UI Upgrades

honeybadgerofdoom:


src/iframe/js/library/charting/chartBtnNewChartWindow.js
```diff
@@ -12,34 +12,38 @@ function showGraph() {
 
 function createAddChartArea() {
```
pierce-smith1, line 13: This function is getting pretty bloated, think about breaking it up

pierce-smith1 (approved):
Other than that one function everything looks great

src/iframe/js/library/charting/chartBtnNewChartWindow.js
honeybadgerofdoom: Yeah that's a good idea, I'll do that real quick

honeybadgerofdoom:
@pierce-smith1 I broke that beastie up

PR #234, opened 2021-03-21
Added Slding Window RPC

drg101:


pierce-smith1 (approved):
Response is messed up but that's not our problem:tm:

PR #242, opened 2021-03-26
https GRPC Route

drg101:


Marc-9 (approved):
Cool we have out own subdomain, but why are we hard coding the port?
Works on my local Firefox

PR #253, opened 2021-04-06
Fix large numeric datasets breaking KDE

### LINKED ISSUE #252: Large numeric datasets cause KDE in histograms to break

pierce-smith1:
Datasets with large individual entries, notably median household income, should no longer cause KDE in histograms to stop working.
Closes #252 .


drg101 (approved):
Works

PR #255, opened 2021-04-13
Bug Fixes

drg101:
Found a bad bug which broke all loading. Also fixed the double reload on worker files.


pierce-smith1 (approved):
Best I can tell nothing is broken.

PR #257, opened 2021-04-14
Defensive optimization

drg101:


pierce-smith1:
It does Work:tm:, but some comments: 
- The cutoff for tracts seems a tad bit overaggressive - it could be zoomed out slightly (though knowing how coarse leaflet zoom levels are, not sure you can really make any minor adjustments to it)
- If you enable tracts at too far out of a zoom level, there seems to be extra latency between inputting a scroll and it actually scrolling in. Have you experienced this / do you know what this might be?
- Very minor formatting problems in the message - not every word should be capitalized, there's no period at the end.
That's all, the popup looks really nice.

pierce-smith1 (changes requested):
see comment

drg101:
@pierce-smith1 I addressed 1 and 3, I can't seem to replicate the zoom latency, but whenever I have run into that issue before though, so we should keep looking into it.

pierce-smith1 (approved):
Tract cap feels a lot better now - message is much better now too.
The zoom level issue feels like a deeper problem, we should make a GH issue for it - this can be merged in for now

PR #261, opened 2021-04-15
Fixed double reloading (probably)

drg101:


drg101:
I just changed the `live-server` config to make it only care about `index.html` and `/dist`

drg101:
I also added alot of things to `devDependencies` that should be there.

pierce-smith1:
Wow, it actually seems to be working. Tried changing js library files, css files, and index.js itself - all changes showed up after one reload. Praise Daniel.

pierce-smith1 (approved):

PR #264, opened 2021-04-16
COVID data plotting

### LINKED ISSUE #263: Implement COVID charting

pierce-smith1:
Closes #263 and #258.

Open the graphing window and create a line graph to see the plots. Should gather data for all counties in the viewport - hovering over a particular line should tell what county it is representing. Data type can be switched between cases and mortality. The window size cannot be changed yet.

This has necessitated a lot of changes to the charting backend, which were previously written under the assumption the only data we would ever be dealing with would come from the map. Please make sure the other charts haven't broken as a result of any of these changes.

Including Matt as a reviewer since the chart creation code he wrote has significantly changed. (Dropdown controls are now class-based: see chartBtnControls, chartBtnCreateChart, controlDropdown, featureDropdown)

Also, for the sake of the servers, respect the defensive optimizations and don't zoom out too much...


drg101 (approved):
Works for me and looks to be doing a minimal amount of intersection queries.

PR #267, opened 2021-04-16
Categorical sliders

drg101:
I also folded in a change to make the defensive stuff come up on the right, because it could be obscured by the menu on the left.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "build": "webpack --mode 'production'",
     "buildDev": "webpack --mode 'development'",
     "watch": "webpack --mode 'development' --watch",
-    "serve": "npm run watch & live-server --ignore='**/node_modules/**,**/src/**,**/index.js/**'",
+    "serve": "npm run watch & live-server --ignore='**/node_modules/**,**/src/**,index.js'",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
     "build": "webpack --mode 'production'",
     "buildDev": "webpack --mode 'development'",
     "watch": "webpack --mode 'development' --watch",
-    "serve": "npm run watch & live-server --ignore='**/node_modules/**,**/src/**,**/index.js/**'",
+    "serve": "npm run watch & live-server --ignore='**/node_modules/**,**/src/**,index.js'",
```
drg101, line 10: Also, the only file which still can have the double reload issue is `index.js` so I'm trying this change out.

pierce-smith1:
One UI comment: is it possible to place the current range string (e.g. "Relatively Moderate - Very High") on its own line? The interface starts looking kind of awkward when this happens:
image
I don't have any major comments on the format of the slider specification from a charting side, it should work fine to piggyback off of what you put in for the categorical plots.

drg101:
@pierce-smith1 Addressed the issue, this actually made a lot of the UI look nicer.

pierce-smith1 (approved):
Nice

PR #270, opened 2021-04-18
Added Unlimited Upper Bound Option to Constraints

drg101:
Now, if a range constraint has `plus=true`, and it's upper bound is selected, the UI will add a `+` after the final value, and `2147483647` will be the upper bound in the query.

### DIFF
--- a/src/js/library/autoMenu.js
+++ b/src/js/library/autoMenu.js
@@ -251,13 +251,15 @@ export default {
                 ...result
             }
             result.type = "slider";
-
+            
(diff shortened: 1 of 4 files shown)


drg101:
Made the label on sliders bold and added unicode arrow cause it looks cool.

pierce-smith1 (approved):
Looks professional AF.
There's a minor issue where, when the slider dropdowns are first opened, the blue filled-in part of the slider doesn't render - at least on my end, anyway (Chrome). May want to look at that eventually.

PR #271, opened 2021-04-19
Remove Blocker Warning if Layer Unchecked

drg101:
Fixed a bug where if you were zoomed out far enough, for a defensive optimization warning to come up, and then unchecked the layer causing it, the warning would persist. This involved making the flag which sets which blockers are on into integers, because the user may select multiple layers.


pierce-smith1:
Did not fix the problem for me - zooming very far out with a bunch of counties and tracts checked and then unchecking all of them still has the warning persist.

pierce-smith1 (approved):
Yeah nevermind, apparently it's fine

PR #272, opened 2021-04-20
Paired Datasets

drg101:


drg101:
@pierce-smith1 Idk whats going on with median county age, but its not client facing so you can still approve.

pierce-smith1 (approved):
County age and county income can't be plotted yet, but that's my problem to fix.

PR #274, opened 2021-04-20
Adjust risk box height

honeybadgerofdoom:
This PR sets the max height for the layer constraint boxes to double their previous value. Also checkbox container max heights are increased.


honeybadgerofdoom:
oops i pulled out of main

PR #276, opened 2021-04-20
Increase max-height

honeybadgerofdoom:
layerConstraints and contentSection classes have their max-heights significantly increased.


honeybadgerofdoom:
Code Climate is pissed

PR #292, opened 2021-04-23
Intersections on Tracts and Counties

drg101:
Please test this a lot, there is probably something wrong with it.

### DIFF
--- a/src/js/library/autoQuery.js
+++ b/src/js/library/autoQuery.js
@@ -535,9 +535,9 @@ export default class AutoQuery {
                 return this.colorCode;
             case "gradient":
                 const range = this.getConstraintMetadata(this.color.variable).range;
-                const normalizedValue = (value - range[0]) / (range[1] - range[0]);
+                const normalizedValue = Math.min(Math.max((value - range[0]) / (range[1] - range[0]), 0), 0.9999999);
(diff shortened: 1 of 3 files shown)


pierce-smith1 (approved):
This is surprisingly robust - playing with it for ~20 minutes or so, querying loads of counties and tracts at the same time revealed very few problems.
The one big problem I was able to find is that other shapes (e.g. flood zones) sometimes show up in front of the topmost tract shape and sometimes under it. Ideally they would intersect as well, but I have no idea how you'd do that.
image

pierce-smith1:
Also, playing with this revealed a charting bug(?) where data being retrieved from the filter doesn't care at all about constraints, so even if you're only looking at two tracts because of specific constraints you'll still get the data for all of them. Will make a separate issue for that.

drg101:
> This is surprisingly robust - playing with it for ~20 minutes or so, querying loads of counties and tracts at the same time revealed very few problems.
> The one big problem I was able to find is that other shapes (e.g. flood zones) sometimes show up in front of the topmost tract shape and sometimes under it. Ideally they would intersect as well, but I have no idea how you'd do that.
> image

Yea this is gonna be tough, I wan't to rewrite the renderer all-together so I'll note this as something to add.

PR #298, opened 2021-04-26
Current Location and Searchbar

drg101:
Added current location and search bar. If you don't allow current location access, the current location icon won't show up.


pierce-smith1:
Works wonderfully. Two oddities:
* The searchbar should really be in the very top-right corner, else it looks silly obscuring so much of the map

* Leaflet flying lags the hell out of my browser (this might be on my end)

PR #299, opened 2021-04-27
Charting window resizable from all four corners

### LINKED ISSUE #295: Make charting window resizable from all four corners (for real)
Didn't think this was necessary, but workshop proved otherwise.

pierce-smith1:
Closes #295.

### DIFF
--- a/src/css/charts.css
+++ b/src/css/charts.css
@@ -76,9 +76,7 @@
 .overlay .option{
     width: 15px;
     height: 15px;
-    border-radius: 50%;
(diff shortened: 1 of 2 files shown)


drg101:
So do all the controls just change where the top right is?

pierce-smith1:
They're not supposed to, it should make the window resize like you'd expect - dragging any corner pulls out that corner and that corner alone

pierce-smith1:
Found a very sloppy bug and fixed it

PR #301, opened 2021-04-28
Dynamic menu sizing

honeybadgerofdoom:
This is an attempt to resize the menu text on different screen sizes. Ideally, someone with a laptop would pull this and see what it looks like on that screen. I can simulate this in chrome by dragging the window size around but it's a little hard to tell what it would actually look like on a laptop

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,6 +6,7 @@ import "./src/css/queries.css";
 import "./src/css/modeling.css";
 import "./src/css/charts.css";
 import "./src/css/main.css";
+import "./src/css/menuResizing.css";
 import "./src/css/bootstrapOverrides.css";
 import "./src/css/third-party/nouislider.min.css";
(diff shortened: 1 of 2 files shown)


pierce-smith1 (commented):
Tried on my laptop, seems to be working great. Though the text does get _uncomfortably_ small for very condensed views - I don't think it should ever go under 14px, niceness of the interface being damned, because it's just way too hard to read.

drg101 (changes requested):
For me at least, anything under 16px makes the menu pretty tough to use. Instead of decreasing the font size as the screen gets smaller, we should decrease the amount of menus we are showing. I think this will be easy when we port the menu over to reactstrap/material-ui. 

Screenshot of how hard it would be to read on a smaller device:
smallText

honeybadgerofdoom:
Oooh I like the idea of just making the menu wider - I'll work on that

PR #305, opened 2021-04-29
COVID QOL and defensive measures

### LINKED ISSUE #296: Block COVID data at low zoom levels

### LINKED ISSUE #302: Change "Line Graph" to "COVID-19" in charting window
Should ideally also find a way to make it clearer to users that COVID data is separate from map data.

### LINKED ISSUE #303: Fix line hovering in COVID chart not working properly

### LINKED ISSUE #304: Add query indicator to COVID chart

pierce-smith1:
Many small improvements to COVID charting.
* Button for the chart says "COVID-19" instead of "Line Graph" (Closes #302)
* Same defensive optimization as regular querying (no queries are possible when very zoomed out) (Closes #296)
  * On top of this, stale queries that take too long are now discarded
* Fixed a problem with line discovery - hovering over lines should be significantly more accurate now (Closes #303)
* The chart will now tell users when it is waiting for data (Closes #304)
* Other very minor improvements
  * The label for a county when hovering over a line should no longer fall off the right side of the screen
  * A subtitle now explains that the data is a sliding average


src/js/library/charting/chartSystem.js
```diff
@@ -224,8 +232,20 @@ export default class ChartSystem {
...
+        this.map.on('zoomstart', e => { 
+            this.doNotUpdate = true;
+        });
+
+        this.map.on('zoomend', e => { 
+            this.doNotUpdate = false; 
+            this.update(); 
+        });
+
+        this.refreshTimer = window.setInterval(() => {
```
drg101, line 246: Why is it re-querying every 2 seconds? I was experiencing some lag on the client because of this

src/js/library/charting/chartSystem.js
```diff
@@ -224,8 +232,20 @@ export default class ChartSystem {
     }
 
     initializeUpdateHooks() {
-        this.map.on('move', (e) => { this.update(); });
-        this.refreshTimer = window.setInterval(() => { this.update(); }, 2000);
+        this.map.on('move', e => { this.update(); });
```
drg101, line 235: `moveend` is a more appropriate event here so it doesn't spam requests like crazy while moving map around

drg101 (changes requested):
Everything seems fine except for the 2 comments I put.

pierce-smith1:
@drg101 I'd like to change the 2-second querying to not happen but it is a little more involved than can be fixed in this PR, so I'll have to make it a separate issue. For now I just removed the 2 second timer completely. This makes the graphs less responsive because you'll need to resize the window yourself to force them to update.

drg101 (approved):
Works

PR #312, opened 2021-05-01
Dark mode charts + resizing fixes

### LINKED ISSUE #307: Fix charting components resizing to disallowed sizes

pierce-smith1:
All charts should now respect dark mode (#308). In doing this, I also had to fix #307, so you get it for free.
Also, lag on the client should be reduced, especially in COVID charts, since queries no longer happen every 2 seconds. Now that timer runs every 1.5 seconds and only does re-rendering of chart frames, nothing more.

### DIFF
--- a/src/js/library/charting/chartSystem.js
+++ b/src/js/library/charting/chartSystem.js
@@ -163,17 +163,7 @@ export default class ChartSystem {
         this.chartFrames = [];
 
         this.resizable = new resizable(resizable.minimum_width, resizable.minimum_height + 200, "white");
-        this.resizable.setResizeCallback((width, height) => {
-            this.chartFrames.forEach(frame => {
-                let newHeight = height - 200;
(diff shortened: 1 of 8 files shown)


src/js/library/charting/chartSystem.js
```diff
@@ -232,7 +222,7 @@ export default class ChartSystem {
     }
 
     initializeUpdateHooks() {
-        this.map.on('moveend', e => { 
+        this.map.on('move', e => {
```
pierce-smith1, line 225: I changed this back because, with `moveend`, the charts do not update seamlessly as you drag the map around, which I think would disappoint Shrideep. 
To prevent queries from getting spammed there's an enforced 0.2 second delay between updates, which has existed for a little while and is not new in this PR (so you won't see it in diffs).
Let me know how you feel about this.

drg101 (approved):
Everything works good, 0.2s seems fine and doesn't lag anything out for me.

PR #331, opened 2021-05-20
Centralized Query System

drg101:


pierce-smith1:
This is great. Query code is so much cleaner now, I like it.
Doesn't seem to have broken anything, but I have two questions:
* what is the difference between query granularities? I don't think I've encountered this term before
* is it just me, or is unchecking a layer a lot slower than it used to be?

drg101:
@pierce-smith1 I'm gonna make a full docs page about this, but:
- `coarse` granularity gives you the length-3 geohash query of an area. This is what you encountered with the covid queries.
- `fine` will only return areas which intersect the bounds.

The `coarse` version accepts a `geohashBlacklist` parameter, which will not query areas in those geohashes. It returns a `geohashes` array which can be used to avoid wastefully re-querying areas.

Which datasets are slow to disappear?

pierce-smith1:
Pretty much all of the county & tract datasets feel slower. Though it could easily be on my end. I'll approve this for now but make a note of it.

pierce-smith1 (approved):

PR #335, opened 2021-06-02
React UI state refactor

honeybadgerofdoom:
Known issue: when you update the workspace, the state of all constraints is reset. I'm still working out the best way to fix this.  The constraints themselves persist, and if you have changed which constraints are active for a given layer that persists as well on workspace update.

This is still being refactored! Definitely a work in progress.

Making this PR to try and avoid/fix merge conflicts going forward.


drg101 (changes requested):
Console errors when updating workspace should be fixed.

honeybadgerofdoom:
@drg101 fixed console errors

honeybadgerofdoom:
Admittedly putting this here mostly for my own reference as I've only just noticed it.

Constraints belonging to layers that are 'higher' (lower index) up in the workspace array persist on workspace update when layer(s) that were added/removed are 'lower' (higher index) on the workspace array. The inverse is not true.

drg101 (changes requested):
I'm noticing a duplicate keys console error on the advanced popup.

drg101 (approved):

PR #341, opened 2021-06-04
Grand Refactoring

honeybadgerofdoom:
This cleans up the code, renames things, creates IndividualConstraint.js, removes ConstraintHelpers.js, adds a tooltip for the optional layer info, and formats the tooltips a bit.


drg101 (approved):
All still works.

PR #343, opened 2021-06-07
Remove Lots of Third Party Code

drg101:
Removed lots of files from `/third-party`. If you see any we don't need please delete them.


pierce-smith1:
* Are we really not using dexie anymore? I thought it was how we were storing the shapefiles?
* Can popper.js also be put on the cutting room floor? I'm unsure where it's used, especially with React now.

drg101:
Dexie is being imported with npm now. @honeybadgerofdoom what about popper?

pierce-smith1:
> Dexie is being imported with npm now. @honeybadgerofdoom what about popper?

Matt should weigh in on this, but a quick search of popper in the codebase shows it's only used in one import statement.

Marc-9 (approved):
Im not using any of those
As long as the site still functions this is fine

pierce-smith1:
FYI, the site does still work on my end, best I can tell.

PR #344, opened 2021-06-08
General Temporal Queries

drg101:
Added:

- Ability to have temporal datasets visualized without randomly returning a time.
- Ability to block a constraint from existing with the `remove` field.
- Ability to set a constraint as the "main" temporal slider with the `temporal` field. This should be in epoch millis.
- Added verbose 'step' property, which can be either `day` or `year` right now, we will add more as need be.
- Added date picker. It's kinda buggy but I didn't want to upgrade to materialUI/next yet. It will be fixed when we switch over to that.
- Ability to make a constraint aggregated with the `temporalType` field. It can be any [MongoDB Accumulator operator](https://docs.mongodb.com/manual/reference/operator/aggregation/group/#accumulator-operator).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3296,6 +3296,22 @@
       "resolved": "https://registry.npmjs.org/@date-io/core/-/core-1.3.13.tgz",
       "integrity": "sha512-AlEKV7TxjeK+jxWVKcCFrfYAk8spX9aCyiToFIiLPtfQbsjmRGLIhb5VZgptQcJdHtLXo7+m0DuurwFgUToQuA=="
     },
+    "@date-io/date-fns": {
+      "version": "1.3.13",
+      "resolved": "https://registry.npmjs.org/@date-io/date-fns/-/date-fns-1.3.13.tgz",
+      "integrity": "sha512-yXxGzcRUPcogiMj58wVgFjc9qUYrCnnU9eLcyNbsQCmae4jPuZCDoIBR21j8ZURsM7GRtU62VOw5yNd4dDHunA==",
+      "requires": {
+        "@date-io/core": "^1.3.13"
+      }
+    },
+    "@date-io/moment": {
+      "version": "1.3.13",
+      "resolved": "https://registry.npmjs.org/@date-io/moment/-/moment-1.3.13.tgz",
+      "integrity": "sha512-3kJYusJtQuOIxq6byZlzAHoW/18iExJer9qfRF5DyyzdAk074seTuJfdofjz4RFfTd/Idk8WylOQpWtERqvFuQ==",
+      "requires": {
+        "@date-io/core": "^1.3.13"
+      }
+    },
     "@discoveryjs/json-ext": {
       "version": "0.5.2",
       "resolved": "https://registry.npmjs.org/@discoveryjs/json-ext/-/json-ext-0.5.2.tgz",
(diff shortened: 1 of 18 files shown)


pierce-smith1 (commented):
image
Some classic javascript right here that should probably get fixed

pierce-smith1:
image
Also this happens, don't know if that will be fixed by the change to materialUI/next

pierce-smith1:
All of these constraints are per-day, right? i.e. the "Max New Cases" is max new cases within a single day, over the time period you requested? In any case I feel exactly what's _going on_ in the queries is not super well communicated to the user. Though I recognize that's not really the point of this PR, I still wanted to bring it up.

drg101:
> image
> Some classic javascript right here that should probably get fixed

I'd fix this, but we're about to rework how this data is displayed anyways.

drg101:
> image
> Also this happens, don't know if that will be fixed by the change to materialUI/next

How did you make this happen? @pierce-smith1

drg101:
> All of these constraints are per-day, right (because "step" is "day" in the metadata catalog for `covid_county`)? i.e. the "Max New Cases" is max new cases within a single day, over the time period you requested? In any case I feel exactly what's _going on_ in the queries is not super well communicated to the user. Though I recognize that's not really the point of this PR, I still wanted to bring it up.

Let me change the labels to clear things up.

pierce-smith1:
> > image
> > Also this happens, don't know if that will be fixed by the change to materialUI/next
> 
> How did you make this happen? @pierce-smith1

By making the max date smaller than the min date. Obviously something should show up to warn me that's not valid, but it looks like the warning isn't displaying right.

drg101:
> > > image
> > > Also this happens, don't know if that will be fixed by the change to materialUI/next
> > 
> > 
> > How did you make this happen? @pierce-smith1
> 
> By making the max date smaller than the min date. Obviously something should show up to warn me that's not valid, but it looks like the warning isn't displaying right.

Can't seem the replicate this. The 'Min Date' on the max date selector gets set to whatever the min date selector is set to.

drg101:
> > All of these constraints are per-day, right (because "step" is "day" in the metadata catalog for `covid_county`)? i.e. the "Max New Cases" is max new cases within a single day, over the time period you requested? In any case I feel exactly what's _going on_ in the queries is not super well communicated to the user. Though I recognize that's not really the point of this PR, I still wanted to bring it up.
> 
> Let me change the labels to clear things up.

@pierce-smith1 Changed the labels to make it make a bit more sense. Really I'd like to let the user create their own indexes based on the accumulator functions, and we'd only provide a few which make sense to everyone. These were more to prove that it works.

pierce-smith1:
> > 
> > By making the max date smaller than the min date. Obviously something should show up to warn me that's not valid, but it looks like the warning isn't displaying right.
> 
> Can't seem the replicate this. The 'Min Date' on the max date selector gets set to whatever the min date selector is set to.

The exact steps are: (Firefox 89.0)
* open menu
* add Covid County to workspace
* enable Covid County
* open constraint dropdown
* click on Max Date
* click 2020 -> Jan -> OK

This is probably just a bug in whatever date selector you're using right now. It _might_ also be a Firefox only issue if you can't reproduce with these steps.

pierce-smith1:
> @pierce-smith1 Changed the labels to make it make a bit more sense. Really I'd like to let the user create their own indexes based on the accumulator functions, and we'd only provide a few which make sense to everyone. These were more to prove that it works.

Right, which I why I said it wasn't really this PR's concern. They do make way more sense now though, so this is good.

drg101:
> This is probably just a bug in whatever date selector you're using right now. It _might_ also be a Firefox only issue if you can't reproduce with these steps.

@pierce-smith1 Ah, this is with that same bug that makes the date selector buggy in the first place. This will be fixed when we switch to the MUI 5.0 date selectors.

pierce-smith1 (approved):

PR #345, opened 2021-06-10
Little UI details

honeybadgerofdoom:


pierce-smith1:
Like the changes - a few comments about the current state of the UI:
* Do we think the menu should be less transparent? As it stands the map bleeds through a _lot_, which is kind of distracting.
* Should the menu close when graphing is opened? I think users would want it to stay open so they can fiddle with the settings while graphing.
* The shadows on each of the container Papers (i.e. the Paper the contains the Advanced, Reset Constraints, etc. controls) are pretty overbearing - they look nice but could be toned down.
  * On that same note, there's a decent amount of wasted space in these Papers. Maybe the margins/padding could be reduced?
* The labels on the constraint sliders (e.g. Heatwave Temperature 86 - 89) wiggle around annoying when you move the sliders, due to the numbers changing length as they are changed. Perhaps the numbers could use a monospace font, or be in their own separate element?

I'll leave these changes at everyone else's discretion - I will approve this PR once the CC issues are fixed.

drg101:
> Like the changes - a few comments about the current state of the UI:
> 
> * Do we think the menu should be less transparent? As it stands the map bleeds through a _lot_, which is kind of distracting.
> * Should the menu close when graphing is opened? I think users would want it to stay open so they can fiddle with the settings while graphing.
> * The shadows on each of the container Papers (i.e. the Paper the contains the Advanced, Reset Constraints, etc. controls) are pretty overbearing - they look nice but could be toned down.
>   
>   * On that same note, there's a decent amount of wasted space in these Papers. Maybe the margins/padding could be reduced?
> * The labels on the constraint sliders (e.g. Heatwave Temperature 86 - 89) wiggle around annoying when you move the sliders, due to the numbers changing length as they are changed. Perhaps the numbers could use a monospace font, or be in their own separate element?
> 
> I'll leave these changes at everyone else's discretion - I will approve this PR once the CC issues are fixed.

+1 On all of these. Try doing 'Power Transmission Lines' while the menu is open, the current opacity makes it pretty hard to read. Will approve if you want to assess these issues in another PR.

honeybadgerofdoom:
@pierce-smith1 @drg101 thanks for the feedback - I like all of these ideas. I'd like to get these things handled in this PR so I'll keep working on it and re-request reviews once I've got them handled.

drg101 (changes requested):
Getting this console error when adding a new layer to the workspace: `react_devtools_backend.js:2560 Warning: Failed prop type: Invalid prop `step` of type `string` supplied to `ForwardRef(Slider)`, expected `number`.`

pierce-smith1:
Everything looks a lot nicer! But I am getting the same problem as Daniel, plus something wack is going on with dates and I'm not sure if this is expected or not:
image

honeybadgerofdoom:
@pierce-smith1 @drg101 those are both related to whatever changes someone else made to master since I've had this branch out - if you look at the Date Range for say, Heatwaves in Master it is taking date range input, whereas my branch is expecting a slider....hence the chaos.

honeybadgerofdoom:
I'll have to work with @drg101 on these things and sort out what changed, then it should be good to go

pierce-smith1:
Cool, I can approve this once the console errors get sorted out.

drg101:
@honeybadgerofdoom It looks like the new version of `IndividualConstraint.js` got replaced (mostly) by the old one. If you just copy the one from master over it _should_ work.

drg101 (approved):
Works

PR #347, opened 2021-06-18
More neon

drg101:


pierce-smith1 (commented):
Seems to work:tm:, but:
I don't know anything about these Neon sets (or what Neon is really), but are they supposed to be so sparse? Some of them don't even have any points at all in certain states.
Also, maybe you have a good reason for this I'm not aware of, but some of the names for these datasets are fairly clunky (e.g. "Asp Air"). Maybe an actual census researcher would be able to understand this better, but I think more descriptive names and/or information tooltips would go miles here.

drg101:
> Seems to work™️, but:
> I don't know anything about these Neon sets (or what Neon is really), but are they supposed to be so sparse? Some of them don't even have any points at all in certain states.
> Also, maybe you have a good reason for this I'm not aware of, but some of the names for these datasets are fairly clunky (e.g. "Asp Air"). Maybe an actual census researcher would be able to understand this better, but I think more descriptive names and/or information tooltips would go miles here.

@pierce-smith1 The sparsity is expected, as there are only ~70 sites total. I will fix all these things in the next PR relating to neon, I will be changing a few things relating to tooltips and such, as some more data is necessary for these datasets.

pierce-smith1 (approved):

PR #348, opened 2021-06-19
Add KDE to ported histogram

pierce-smith1:
This re-introduces KDE to histograms in a React-friendly way.
Known issue: the Input up/down buttons don't work on Firefox. No idea why - I think it honestly might be a bug in MaterialUI.


Marc-9:
KDE works as intended
Fixes with graph visibility need to be fixed before this is pushed to master
Cant test the Firefox bug as Aperture just does not load for me in Firefox at all

PR #351, opened 2021-06-21
Ported charts

Marc-9:
Adds support for many new graphs
Histogram
ScatterPlot
Line Chart (covid only atm)
Box Plot (under name pie graph)

It also re-organizes the structure of charting and deletes lots of old unnecessary charting files

Fixes
- Better formatting (margins, font)
- Scroll using scroll bar
- Add more buttons (or change design for new graphs)
- Don't allow scatterplot of tract vs county
- Risk Tract does not show up in dropdown
- Box Plot needs dynamic x axis
- Remove individual graphs

New graph types (for the future)
- Radar Chart (for comparing counties/tracts over multiple constraints)
- Correlogram (view correlations between constraints)
- Ridgeline (visualize values over time frame)


pierce-smith1 (approved):
Works:tm:.

This PR is big enough, but some more things we need to fix in the future (and by 'we' I do actually mean 'we', since some of these should be my responsibility):

* the "Select Constraint" boxes for scatterplots need to specify which is x and which is y
* the KDE controls for histograms are unlabelled so it's not obvious what they are
    * these controls should also probably be hidden until someone actually asks for KDE, to reduce clutter
* the ability to hover over lines in the covid chart to tell what county they're from should be restored
* scatterplots still use disgusting internal Feature names instead of human-readable ones

If you agree we could make Github issues for each of these and distribute responsibilities.

Not totally sure what to do about CC: a lot of the issues are invalid because React functional components kind of have to be longer than 25 lines. Advice? (Do we just wontfix everything?)

Edit: removed some complaints because you already mentioned them in the PR body.

drg101 (dismissed):
errors

Got these errors by trying to use the histograms. I also couldn't seem to get the option to make a county svi histogram to show up. The first one seems like an issue worth fixing. The warning happens whenever you remove a layer from the map while charting one of its features. I'm not sure what causes the useEffect error.

@pierce-smith1 I set the line-count issues to `invalid` since they are a bit silly for functional components. I left the complexity ones open, but you guys can close them if you feel the component is simple enough.

I will approve once there are no errors/warnings in the console with normal use.

pierce-smith1:
> I also couldn't seem to get the option to make a county svi histogram to show up. 

I can reproduce the warnings/errors when a layer is removed, but can't reproduce this in either ff or chrome - can you give a set of steps?

drg101:
> > I also couldn't seem to get the option to make a county svi histogram to show up.
> 
> I can reproduce the warnings/errors when a layer is removed, but can't reproduce this in either ff or chrome - can you give a set of steps?

Well now that seems to be working for me, weird.

Marc-9:
There should no longer be warnings or errors in the console, the above fixes by Me and pierce should still be noted and worked on in the next branch

pierce-smith1:
Can confirm this issue does appear to have been fixed, at least on firefox.

drg101 (commented):
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    at ChartingResizable (webpack://aperture/./src/js/ui/NewCharting/ChartingResizable.js?:24:72)
    at div
```
To Recreate:

1. Add heatwaves to map
2. Open graphing
3. Add histogram
4. Move around a bit so that the graph is affected by the new data
5. Close the graphing popup 

I guess the `useEffect` listeners are still registered when the reliable disappears for some reason. This **might** be a race condition cause I can't get it to always happen.

If you guys want this in master really badly, I guess we can merge it, but this issue should be fixed ASAP.

drg101:
> I guess the useEffect listeners are still registered when the reliable disappears for some reason. This might be a race condition cause I can't get it to always happen.

For some reason, the warning line is on the `setSize`, not sure why this is. I am on chrome btw.

pierce-smith1:
> This **might** be a race condition cause I can't get it to always happen.

This **is** a race condition. If the chartSystem is in the middle of processing data and a resizable unregisters itself, it can't always see the unregister in time before it passes the data. 

I know how to fix this, but I need to leave for the dentist in 20 minutes so I'll be MIA for a while. I'll address this when I come back.

pierce-smith1:
@drg101 I think I've addressed the problem - A race condition might still be possible but it will at least be significantly rarer now - can you see if you can still reproduce it?

drg101 (approved):
Looks like a race condition *should* be impossible now.

PR #356, opened 2021-06-24
New date selector 2

honeybadgerofdoom:


drg101 (changes requested):
A few things:

- The menu can be crashed by removing all text in a picker
- The max time can be set before the min time, and vice-versa. Use the min/maxDate property to fix this. It is already set up for the other date picker.

drg101 (approved):
Wrong PR oops.

drg101 (changes requested):
View original review

pierce-smith1 (commented):
Will approve once Daniel's comments are fixed

Marc-9:
Can you not select a day?
For neon 2d wind (safari) I can select a year and month but have no option for day

honeybadgerofdoom:
@drg101 I am not able to set the min date before the max or vice-versa, and this branch does use the same minDate/maxDate logic as yours does. Which dataset are you noticing this for? Am working on the crash bug rn.

honeybadgerofdoom:
@drg101 @pierce-smith1 those 2 bugs are fixed

honeybadgerofdoom:
@Marc-9 some datasets have finer/coarser temporal resolution than others. Heatwaves, for example,  can only select year whereas many NEON datasets can select minutes.

drg101 (approved):
Works for me.

pierce-smith1 (approved):

PR #358, opened 2021-06-25
Better Popup

drg101:


pierce-smith1:
Works:tm:, but:

Some MongoDB is leaking through: Most datasets have something like this
image

Covid county has a particular problem with dates
image

On neon barometric pressure: if a measurement doesn't have units, it probably shouldn't display anything, the NA here is very confusing
image

I would consider the state and county/tract name to be important fields, especially since it's hard to tell right now what area the popup is even showing data for - maybe these should be added up here? (The state code should be resolved to a real state name ofc)
image

I'm not sure how you're supposed to gracefully solve the mongodb object problem - maybe you can try to decode the value as an object and if it's a simple key-value pair, just show the value? Not sure. You can solve this in another PR. I'd like to see at least the state and county/tract name in the important fields, or some other way to tell what area you're looking at data for.

pierce-smith1 (changes requested):
See above

pierce-smith1:
Also I just noticed that **Fahrenheit** is misspelled on any UI component that adds it as a unit

drg101:
@pierce-smith1 Addressed all of your issues, thanks for the good ideas. 

- Added ability to make a field "important" by default. For right now, this is done globally in a JSON file, this can get a bit wonky if there are two "State" fields, and then they both show up on the important section. I'll fix this later. I've added state, county, city, name, and some other things to this file.
- The state code gets converted to a name.
- $numberLong and $numberDecimal are now turned into numbers. If we encounter more rogue mongo objects, we can easily extend the logic that implements this.
-  Fixed Fahrenheit typo.

pierce-smith1 (approved):
Wow, can't say I expected you to just fix everything I mentioned. This is awesome.

I was able to notice one other wild MongoDB object, which appears on county SVI (and probably some others):
image

Given that this is an internal object it probably just shouldn't show up at all. But I'll leave fixing that at your discretion for this PR, I'm more than happy to approve this in its current state.

drg101:
@pierce-smith1 I added `id` and `_id` to the list of blocked keys.

PR #360, opened 2021-06-28
Temporal Charts on Popup

drg101:
Added temporal charts to popups. I used nivo because it seems like our version of the chart popup was more focused around covid and I wanted to get this prototyped quickly. Once ours matches nivo's capabilities we can switch over.

Some of the code is a bit wonky, cause this implementation should just be temporary.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3680,11 +3680,421 @@
         "react-is": "^16.8.0 || ^17.0.0"
       }
     },
+    "@nivo/annotations": {
+      "version": "0.71.0",
+      "resolved": "https://registry.npmjs.org/@nivo/annotations/-/annotations-0.71.0.tgz",
+      "integrity": "sha512-/72r7DmHJeisYTfX+xIiDefPN05N/ItW3HvPO1niTYH2xlnhnStlkAU9pvbRH2Nlxv01/nZWgtv9zJFQ/RW6uw==",
+      "requires": {
+        "@nivo/colors": "0.71.0",
+        "@react-spring/web": "9.2.0",
+        "lodash": "^4.17.21"
+      }
+    },
+    "@nivo/axes": {
+      "version": "0.71.0",
+      "resolved": "https://registry.npmjs.org/@nivo/axes/-/axes-0.71.0.tgz",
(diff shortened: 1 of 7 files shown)


pierce-smith1 (commented):
COVID looks really good, but what's going on with heatwaves?
image

drg101:
@pierce-smith1 Nothing is wrong with the graph, but the underlying data has something fundamentally wrong with the way it is stored. Apparently there is multiple records for each year, but the timestamp is always just the year. 

The graph doesn't seem to play well with multiple entries on the same date, but this isn't a problem for correctly formatted data.

pierce-smith1 (approved):
Approving on the condition that nivo will eventually be removed once our in-house line charts become more timeseries general

Marc-9:
Very cool, I believe you mentioned this earlier but you want the graphs to be made in such a way so that they can be pulled and placed within the popup just like the naive graphs you have here?
I would like a way to close the popup other than clicking on the X, I like clicking on the map but perhaps a persistent tab closer that follows you as you scroll down

Marc-9 (approved):

Marc-9 (changes requested):
Why is it if I select tract median household income then tract median age, when I click on a tract the median household income is included in important info, but If I deselect then both and reselect them in the other order tract median age is included in important information but not the median income.
Can we not have both of their data in important info if they are both selected?

pierce-smith1:
> Why is it if I select tract median household income then tract median age, when I click on a tract the median household income is included in important info, but If I deselect then both and reselect them in the other order tract median age is included in important information but not the median income.
> Can we not have both of their data in important info if they are both selected?

That's a good catch, the popup doesn't appear to be handling unions correctly - tested it with a couple other county/tract datasets, and it seems whichever you select last is what goes into the popup, with data from the other set not appearing the important info fields.
Though I think that should be addressed in a separate PR, since this one is about graphing.

drg101:
@Marc-9 @pierce-smith1 Yea I'm only pulling the "important" fields for the most recent addition to the union. I'll fix this bug in the next PR.

drg101:
> Very cool, I believe you mentioned this earlier but you want the graphs to be made in such a way so that they can be pulled and placed within the popup just like the naive graphs you have here?
> I would like a way to close the popup other than clicking on the X, I like clicking on the map but perhaps a persistent tab closer that follows you as you scroll down

I like the idea of it being sticky to the top, I will also implement this in the next PR.

PR #362, opened 2021-06-29
Reimplement covid line tooltips

### LINKED ISSUE #361: Re-implement interactivity to covid line chart

pierce-smith1:
Returns the old functionality where you could hover over lines in the COVID chart to see what county there were a part of.

The Rnd component made this an _absolutely colossal_ pain, for more reasons than one. Please test thoroughly, as this fix definitely involves some really hacky duct-tape solutions.

Closes #361.

### DIFF
--- a/src/js/ui/NewCharting/ChartingResizable.js
+++ b/src/js/ui/NewCharting/ChartingResizable.js
@@ -14,6 +14,7 @@ function shouldAvoidDragging(node) {
 
 export default function ChartingResizable() {
     let [size, setSize] = useState({ width: 500, height: 400 });
+    let [pos, setPos] = useState({ x: 900, y: 100 });
     let [chartData, setChartData] = useState({});
 
     useEffect(() => {
@@ -46,10 +47,14 @@ export default function ChartingResizable() {
                 onDrag={(node, x, y, deltaX, deltaY, lastX, lastY) => {
                     return !shouldAvoidDragging(node.target);
                 }}
+                onDragStop={(node, x, y, deltaX, deltaY, lastX, lastY) => {
+                    // Yes, this is not a mistake, the library is bugged.
+                    setPos({ x: x.x, y: x.y });
+                }}
             >
                 <Paper className={'charting-resizable-window'}>
                     <div style={{ overflowY: "scroll", maxHeight: size.height }}>
-                        <ChartingWindow size={size} data={chartData}/>
+                        <ChartingWindow size={size} pos={pos} data={chartData}/>
                     </div>
                 </Paper>
             </Rnd>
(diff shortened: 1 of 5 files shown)


src/js/ui/NewCharting/ChartingResizable.js
```diff
@@ -46,10 +47,14 @@ export default function ChartingResizable() {
                 onDrag={(node, x, y, deltaX, deltaY, lastX, lastY) => {
                     return !shouldAvoidDragging(node.target);
                 }}
+                onDragStop={(node, x, y, deltaX, deltaY, lastX, lastY) => {
+                    // Yes, this is not a mistake, the library is bugged.
+                    setPos({ x: x.x, y: x.y });
```
drg101, line 52: Legendary piece of code right here.

drg101 (changes requested):
A few things:
The hover-over seems inaccurate, watch this video:
https://streamable.com/cn3azk

The [cursor](w3schools.com/cssref/pr_class_cursor.asp) shouldn't be a "move" cursor whenever mousing over the graphs. Maybe we should implement it like windows and only allow the moving to be done on the header.

The tooltip seems to be a fixed distance from the cursor up-and-to-the-left. This means it disappears if you go too far to the left side of the graph, or too far up.

On a separate note, should the hover give the value of the y and x-axis too? This will probably be a pain to implement, but it is needed to match most pre-built charting framworks.

pierce-smith1:
> The hover-over seems inaccurate, watch this video:
Unfortunately, what I'm seeing in that video is the cursor transformation that took so much effort to make not working... I have to wonder if it's a cross-browser thing, since I haven't tested it yet outside FF. (I'll do that soon)

> The cursor shouldn't be a "move" cursor whenever mousing over the graphs. 
I agree, and I _think_ this can be somewhat easily fixed by some Rnd props.

> The tooltip seems to be a fixed distance from the cursor up-and-to-the-left. 
It's supposed to be right above the cursor, and there is logic to bump it leftwards when you get too close to the right side - it just looks very bad on your end because the cursor transform isn't working. (There is no logic to bump it downwards if it gets too close to the top, but this tends to be a non-issue since the graph very rarely goes that high)

> should the hover give the value of the y and x-axis too?
Yes. 

> This will probably be a pain to implement
Possibly, but not necessarily - I will give it a try in this branch. The hard part of this PR was programmatically locating the cursor between all the breaking crap that Rnd does, drawing new stuff should be (relatively) simple.

pierce-smith1:
@drg101 This should take care of the big issues - adding more details to the pop-up will be done later in the more general fashion we can re-use for other graphs.
There are two CC issues, but they're somehow not related to the line chart at all, and refactoring to fix them would be fairly complicated and outside the scope of this PR.

drg101:
All seems to work.

drg101 (approved):

PR #367, opened 2021-06-30
Popup Bug Fixes and Refactor

drg101:
Fixed all the bugs I have found and refactored a bit.

@Marc-9 I didn't make the 'x' easier to access yet, because I'd like to let @honeybadgerofdoom finalize the styling of this before I put too much styling.

### DIFF
--- a/src/js/library/autoQuery.js
+++ b/src/js/library/autoQuery.js
@@ -326,8 +326,7 @@ export default class AutoQuery {
         //console.log(this.constraintData[this.temporal])
         indexData[this.collection] = {
             "color": this.getColor(data.properties),
-            "joinField": this.data.joinField,
-            "temporalRange": this.constraintData[this.temporal]
+            "joinField": this.data.joinField
         }
 
         data.properties.meta = this.buildMetaMap();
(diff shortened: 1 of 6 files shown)


pierce-smith1 (changes requested):
It does Fix:tm: the union issue, but some more problems I noticed that should be squashed for this bug fix PR:

~~If two sets disagree on how to format state or county, it seems the table just shows both:~~
image

Edit: This problem seems to actually just be happening to raw datasets, no need to even start a union:
image

Looking at tracts (in general, not necessarily over a union) reports the name of the tract as "County":
image

drg101:
@pierce-smith1 I think I fixed all of the issues that you noted.

pierce-smith1 (approved):
LGTM
"Area name" is a tad clumsy but it will work for now.

Marc-9 (approved):
Do you guys think
LGTM = Looks Good To Me
or
LGTM = Lets Get inTo Master

Ive always read it as the ladder despite the fact I know its the former

pierce-smith1:
> Do you guys think
> LGTM = Looks Good To Me
> or
> LGTM = Lets Get inTo Master
> 
> Ive always read it as the ladder despite the fact I know its the former

:sparkles: _why not both_ :sparkles:

PR #368, opened 2021-07-07
Add license agreement to files

Marc-9:
Let me know if I missed any files, or if there are files in here that should not be
Also IntelliJ gave me tons of crud on this push, citing tons of warnings and some "errors" but it was unrelated to what I added lmk if we are just ignoring those or if I should try and fix those


Marc-9:
For some reason it says I deleted lines?
This is a weird GH thing right?

Marc-9:
Also current branches with new files will need the license, or just pm me when you merge into master and I can add them in with the script

pierce-smith1 (commented):
Doesn't break anything:tm:
I'll need Daniel to weigh in on this, but some of the JS files are autogenerated, right? If so then putting comments on them is probably a bit of a waste and you may want to exclude them.
Also, I genuinely can't figure out when the diffs decide that a file has been completely deleted and re-written (causing GH to think there was deletions) or when a file has just had the license added to the top. It seems completely random.

pierce-smith1:
> Also current branches with new files will need the license, or just pm me when you merge into master and I can add them in with the script

We could just require that any new PRs have licenses at the top of the files, to make things less work for you

drg101:
The `src/js/grpc/GRPC_Querier/grpc_querier.bundle.js` should be excluded because it is autogenerated. Everything else looks fine.

drg101 (commented):
See comment

drg101:
> For some reason it says I deleted lines?
> This is a weird GH thing right?

This is just a git thing, it thinks you deleted the top of a lot of files, not just appended to them. idk why

drg101 (approved):

PR #370, opened 2021-07-21
Color Changes

drg101:


drg101:
- You can switch the field which color codes the collection which is rendered.
- Icons have color coding now
- There is a legend which has information about what you are color coding.

drg101:
Please look for bugs, there is likely a few

src/js/ui/hooks/useHover.js
```diff
@@ -0,0 +1,24 @@
+//from https://usehooks.com/useHover/
```
pierce-smith1, line 1: File needs license

src/js/ui/widgets/Popup/PopupTableEntry.js
```diff
@@ -0,0 +1,36 @@
+import React, { useEffect, useState } from "react";
```
pierce-smith1, line 1: File needs license

src/js/ui/widgets/Popup/PopupTableValue.js
```diff
@@ -0,0 +1,32 @@
+import React, { useState } from "react";
```
pierce-smith1, line 1: File needs license

src/js/ui/widgets/Popup/PopupColorInfo.js
```diff
@@ -0,0 +1,154 @@
...
+            svg.append('line')
+                .style("stroke", "black")
+                .style("stroke-width", 3)
+                .attr("x1", xValue)
+                .attr("y1", 0)
+                .attr("x2", xValue)
+                .attr("y2", height - marginBottom);
+
+            //add axis
+            const linearScale = d3.scaleLinear()
```
pierce-smith1, line 112: Coloring based on date is really odd:
image

(I _think_ this is related to the use of `scaleLinear` here: maybe you want to `scaleUTC` when talking about dates? Do we even want people to color code off of date?)

src/js/library/renderInfrastructure.js
```diff
@@ -278,9 +283,27 @@ export default class RenderInfrastructure {
             return {
                 ...acc,
                 ...curr.properties,
-                meta: { ...acc.meta, ...curr.properties.meta }
+                meta: { ...acc.meta, ...curr.properties.meta },
```
pierce-smith1, line 286: Unions aren't colored at all, though the popup kind of implies they should be: 
image

(I _think_ this is related to this part of the code)

src/js/library/Color.js
```diff
@@ -0,0 +1,157 @@
...
+*/
+import Gradient from "../third-party/Gradient"
+
+/**
+ * @class Color
+ * @file Takes some parameters about a field in constructor, has a method for getting the color of a field based on whatever value is present.
+ * @author Daniel Reynolds
+ */
+
+export const defaultGradient = ["#c7445d", "#e07069", "#f0d55d", "#509bc7", "#4d6dbd"];
```
pierce-smith1, line 67: Something to consider: The default gradient assumes that lower = worse and higher = better, which is true for some datasets but not all - this is really confusing on things like heatwaves, where it makes higher temperatures more blue. Maybe datasets should have some notion of which end of the spectrum is "worse", and flip the default gradient accordingly?

src/js/library/renderInfrastructure.js
pierce-smith1: Hospitals are black voids of despair by default:
image

(Maybe also related to this)

src/js/library/Color.js
```diff
@@ -0,0 +1,157 @@
...
+        }
+        else if (predefinedColor.style === "sequential" && predefinedColor.map) {
+            this.colorMapping = predefinedColor.map;
+        }
+        else if (predefinedColor.style === "sequential") {
+            this._createDefaultColorMapping()
+        }
+    }
+
+    _createGradient(gradientArr, resolution = 100) {
```
pierce-smith1, line 133: It's a little strange how gradients with fields that are sometimes missing works: I wouldn't necessarily expect a "missing value" to have a colorful entry in the gradient (maybe it should just be gray)?
image

src/js/ui/widgets/Popup/PopupColorInfo.js
pierce-smith1: The default value that is colored on for some datasets is strange: on Neon datasets it often defaults to coloring by site rather than the actual data of interest, and on covid county for instance, it colors on state: 
image

pierce-smith1 (changes requested):
I tried doing an actual GitHub review for this to organize the comments, and apparently it doesn't let me leave comments without attaching them to some line of source, so you can mostly ignore that.

I think this is one of the coolest features we have on the client now, and it works _really_ well in general (aside from the re rendering slowness)

src/js/library/renderInfrastructure.js
pierce-smith1: This also applies to some neon data points - I think it's just based on values being missing from whatever field it defaults to

drg101:
@pierce-smith1 Yea, I'll add the default field for those collections which pick weird fields by default. The union thing though, I'm gonna leave as is because I'm gonna rewrite how the union colors are programmed.

drg101:
@pierce-smith1  For the gradient direction, I agree, I'll add an attribute for this.

Marc-9 (approved):

Marc-9:
First off this is awesome.

Secondly, we must be close to the mountain top right?  How much more can we put these poor browsers through?

Feature Request - when you select a different field to color code the map on, that field should then be added to the important fields tab.

Bug - When charting is open and a county/tract is selected to open the side menu charting is closed (however when charting is closed all graphs are losted)
However this is a problem I most likely should handle, at the moment there is no way to close the resizable without loosing the graphs, either the frames (graphs) need to be held high up or I can add a minimize feature that you can call when the side bar opens.

drg101:
> Secondly, we must be close to the mountain top right? How much more can we put these poor browsers through?

This is pretty badly optimized right now, so we still got a ways to go :)

drg101:
> Feature Request - when you select a different field to color code the map on, that field should then be added to the important fields tab.

@Marc-9 Done

drg101:
> Bug - When charting is open and a county/tract is selected to open the side menu charting is closed and this causes all graphs within it to be lost. A quick fix would to leave the resizable open and just give the side menu a higher z-index.
> However this is a problem I most likely should handle, at the moment there is no way to close the resizable without losing the graphs, either the frames (graphs) need to be held high up or I can add a minimize feature that you can call when the side bar opens.

Yea Idk what is happening here, I'd be happy to help fix.

drg101:
@pierce-smith1 I've addressed all of the issues I said I'd address.

- Added dateUtc for date fields. I'm gonna leave the color coding for date available because it seems like maybe someone will want to use it.
- Instead of collections choosing a color by default, they now will wait for the user to choose one. By default, they are either black (polys and lines) or white (points)
- Added attribute to flip gradients when higher=worse.
- An empty string will always be grey now.

Marc-9 (changes requested):
Thanks for adding in my change, as for the graph bug, as we discussed over teams by setting the default state in root.js to true for chartingOpen this keeps the graph open even when a popup is opened.
And since the global state of the chart is false inside of setPopupObj this leads me to believe that there is some intermediary state where the default state is being reverted to.
For instance even if you set the sidebar to true in the default state it still gets closed when the popup is opened that's because setPopupObj closes it, which is why I came to the conclusion above.
If one of yall solves this please let me know what the problem was. My assumption is its somewhere in render infrastructure.

drg101:
> Thanks for adding in my change, as for the graph bug, as we discussed over teams by setting the default state in root.js to true for chartingOpen this keeps the graph open even when a popup is opened.
> And since the global state of the chart is false inside of setPopupObj this leads me to believe that there is some intermediary code where the default state is being reverted to.
> For instance even if you set the sidebar to true in the default state it still gets closed when the popup is opened that's because setPopupObj closes it, which is why I came to the conclusion above.
> If one of yall solves this please let me know what the problem was. My assumption is its somewhere in render infrastructure.

@Marc-9  Yea Idk why the default global state is being mixed in on every update, this is actually a pretty big bug so it's worth taking care of. I will try to fix this tonight.

drg101:
@Marc-9 Fixed your issue

pierce-smith1 (approved):
Everything that I mentioned seems to have been fixed.

Marc-9 (approved):

PR #373, opened 2021-08-03
Faster Rendering and Color Changing

drg101:


src/js/library/renderInfrastructure.js
```diff
@@ -120,14 +137,16 @@ export default class RenderInfrastructure {
         if (this.options.simplifyThreshold !== -1) {
             Util.simplifyGeoJSON(geoJsonData, this.options.simplifyThreshold);
         }
-        //console.log(geoJsonData)
 
         Util.fixGeoJSONID(geoJsonData);
+        const iconName = Util.getNameFromGeoJsonFeature(geoJsonData, indexData)
+        geoJsonData.properties.apertureName = iconName;
```
pierce-smith1, line 143: This is causing a test failure - the mock `geoJsonData` object provided by the tests does not have a `.properties.`

pierce-smith1 (changes requested):
I never thought I would see the day the client was _this_ fast. Will approve when you address the missing license, fix the tests and address CC.

drg101:
- Sped up rendering on a general level
- Very fast color switching
- Added an intersection switch onto the data exploration tab. This will make tracts and counties only display if they have fit ALL of the collections, and their corresponding fields/constraints. 
- Known issue/thing: when the intersection tab is on, rendering is a bit clunky, especially with tracts. This is because it is nearly impossible to sync up the rendering processes without destroying speed.

honeybadgerofdoom (approved):
w0w so speedy! Looks good to me.

drg101:
Hold up, found a bug.

pierce-smith1 (approved):
Hard to believe that this works so well.

One minor point is that we could definitely use a help tooltip or something alone those lines for the intersection button, since its use isn't super obvious from the name alone. (There also might be a better place for it in the UI)

Marc-9 (approved):
I thought I found a bug with turning on/off the intersection after adding a new layer but after not being able to recreate it I think the server just happened to go down for a moment in time, as the menu was not showing up for a while after reload.
It also appears that this will fix the multiple layers issue in graphing, so Id love to get this in asap so I may merge master into my branch again and get my PR up and running

PR #375, opened 2021-08-06
Chart Fixes

Marc-9:
Im not sure if code was lost or if master does stuff differently but there are some bug fixes that need to be resolved
1. No more than 1 constraint appears as an option (code is handled in make frame, selected constraint)
2. Deleting a frame make the frame above it inherit its selected constraint


Marc-9:
Issues to resolve before merge
Delete issue- Talked with Matt about this, seems to be a key problem, all frames (graphs) are stored in an array and passed their index in the array, on delete the graph deletes itself from the array, however the graphs after it inherit their selected layer.
Add Delete button on covid date

To be fixed later
Sizing issues- Some graphs don't look great in the resizable this is a sizing issue hopefully Matt can look at
Box Plot ugliness- The box plot is pretty small for some layers but not for other, it also is missing its one of its lines

src/js/ui/NewCharting/ChartTemplates/CorrelogramGraph.js
```diff
@@ -0,0 +1,90 @@
...
+                if(key == key2){
+                    data[counter][key2] = 1;
+                }
+                for (const [key3, value3] of Object.entries(value)){
+                    if(key3 in value2){
+                        x.push(value3)
+                        y.push(value2[key3])
+                    }
+                }
+                const correlation = calculateCorrelation(x, y);
```
pierce-smith1, line 37: This is causing a crash if a tract and county dataset are both enabled at a time while the correlogram is open.

To reproduce:
* enable a county dataset (i.e. heatwaves)
* open the correlogram 
* enable a tract dataset (i.e. tract svi)
* wait for death

src/js/ui/NewCharting/ChartTemplates/CorrelogramGraph.js
```diff
@@ -0,0 +1,90 @@
...
+
+export default function CorrelogramGraph(props) {
+    let retData = {};
+    let keys = [];
+    let data = []
+
+    if(props.data['map_features']) {
+        for (const [key, value] of Object.entries(props.data['map_features'])) {
+            if (value.length > 0) {
+                keys.push(key.split("::")[2])
```
pierce-smith1, line 14: A few more `split`s to replace

pierce-smith1 (changes requested):
I'll approve this once the notes I placed on the code get resolved and the CC problems are either ignored or fixed (the d3 complaints about long functions can be ignored). There are more things we should take note of:

* Histogram bars are black sometimes when they are first created
image

* Scatterplots show erroneous 0 entries for (what I'm assuming are) locations that don't both have data for a particular constraint (e.g. here, I'm looking at counties that don't have heatwave data but do have svi data, and the dots that don't have heatwave data are shoved into 0 degrees)
image

Note that this _only_ occurs when the dataset with missing entries on the y axis.


* The covid graph has lost its ability to hover over counties somehow (great)

* The correlogram still isn't working right? the diagonals should be 1.0 and the off-diagonals should be mirrored, but this correlogram:
image
...looks like 
```
1.0   -0.063
-0.53 0.16
```
instead of something like
```
1.0   -0.53
-0.53 1.0 
```

Out of these the correlogram is really the highest priority, since if the graph is incorrect that's bad news. I can help with the histogram rendering issue and covid graph issue, since those are probably my fault anyway.

drg101:
Screenshot 2021-08-10 6 17 25 PM

An issue with scatterplots when one collection has entries rendering which the other does not.

drg101:
Screenshot 2021-08-10 6 29 08 PM
A lot of things going wrong in this screenshot. Weird data, console logs, console errors.

pierce-smith1:
> +1 On all of pierce's reviews as well, but more than anything else, the whole graphing system feels "off."
> Here is a video of me trying to create a histogram. It doesn't seem to update in real-time anymore.
> https://streamable.com/vmfp4g
> 
> The system feels like this for every other graph as well.

This is very noticeable and as the person behind the original infrastructure I want to do some investigation into why it's happening. I really hope this isn't a problem with React.

Marc-9:
Right now I believe I have fixed the scatterplot issue and correlogram, however I have only set up the correlogram to work with tracts, I reckon I could get them to work with counties but its less useful with less data, and I think Shrideep cares more if it works with the neon data.
I can fix the failing test issue next but as for the black bars and slow re-rendering not sure what I can do in this branch at least

Marc-9:
I still have other things to work out but let me know if you see new things or if the issues I think I resolved were not really fixed

pierce-smith1:
I am still seeing the scatterplot bug: 
image

The correlogram is doing better in terms of accuracy, but the bottom-left squares are completely black and have no value:
image

This in console:
image

Also, if you open the correlogram with nothing being rendered, the client crashes with this in console:
image

Marc-9:
@pierce-smith1 can you give me the location where this scatterplot issue is showing up
As for the bottom left in the correlogram, those are equivalent to the top right, so to save confusion they are just black

Ill work on the console bugs rn

pierce-smith1:
> @pierce-smith1 can you give me the location where this scatterplot issue is showing up
> As for the bottom left in the correlogram, those are equivalent to the top right, so to save confusion they are just black
> 
> Ill work on the console bugs rn

Right in the middle of South Dakota is a region missing a lot of heatwave data, you can recreate the bug fairly reliably there.

I'd argue that making the lower left entries in the correlogram black and valueless is more confusing than just having them mirror across the diagonal, but maybe someone else can provide some input.

Marc-9:
Here is what I believe we have left before merging
1. Correlogram should have no errors in console (but can look ugly atm)
2. Fix the key/indexing bug that causes deleted frames to have their properties inherited by the frame next in the array

I want to finish these today but if I cant get the menu to load it may be a bit later

pierce-smith1 (changes requested):
The scatterplot bug is fixed, but the correlogram is now crashing when any data is all is enabled on the map:
image

Also, creating and deleting chart frames seems to have an extremely long delay. Unless you double-click them, it seems to take anywhere from a couple seconds to forever to actually create or delete the chart. I don't remember this behavior in previous commits of the branch.

Marc-9:
> The scatterplot bug is fixed, but the correlogram is now crashing when any data is all is enabled on the map:
> image
> 
> Also, creating and deleting chart frames seems to have an extremely long delay. Unless you double-click them, it seems to take anywhere from a couple seconds to forever to actually create or delete the chart. I don't remember this behavior in previous commits of the branch.

Yup sorry about that, uncommitted change fixed the correlogram
As for the delete issue, I don't see the same problem on my end, deletes are relatively quick (1-2 seconds)
But it did make me realize a bug in the way I handle graphs, which I will work on solving now, can anyone else see if deletes are causing issues

drg101 (approved):
No errors in the console, but everything still seems sluggish. We can talk about how to resolve this at the next meeting. For now though, this is an improvement.

PR #378, opened 2021-08-09
More Datasets

drg101:


drg101:
- Added many more geometry datasets
- Added a larger proximity dataset
- Added an internet access dataset
- Added more neon datasets

Note: I set all new neon temporal fields to 'avg' because Matt & I are setting up a system where the user can select whatever aggregation parameter they want.

pierce-smith1:
One important issue: 

Private Schools appears twice in the menu - checking/unchecking one will check/uncheck the other
image

Then, when actually enabling Private Schools on the map, the displayed data is for Oil and Natural Gas: 
image

Smaller issues that are up to you to fix:

Display issue in DHL Facilities/PLACEMENT field (also shows up for Fedex Facilities):
image

What is the difference between these two fields in Hydrocarbon Pipelines?
image

pierce-smith1 (changes requested):
See above

drg101:
> One important issue:
> 
> Private Schools appears twice in the menu - checking/unchecking one will check/uncheck the other
> image
> 
> Then, when actually enabling Private Schools on the map, the displayed data is for Oil and Natural Gas:
> image
> 
> Smaller issues that are up to you to fix:
> 
> Display issue in DHL Facilities/PLACEMENT field (also shows up for Fedex Facilities):
> image
> 
> What is the difference between these two fields in Hydrocarbon Pipelines?
> image

@pierce-smith1 
- Fixed the label issue on oil wells.
- Idk whats going on with the \n thing, we can just ignore that since its bad data.
- Blacklisted the Shape_leng(th) fields, along with a few similar others. These were really just artifacts from when these datasets were shapefiles.

pierce-smith1 (approved):
Shape length property still leaking through on the Trails set.
image

The major problem is fixed, though.

One thing to note for a future PR is that unchecking some of the really point-dense datasets can take a long time, so it may be worth looking for optimizations here.

Marc-9 (changes requested):
Bug- 
Zoom out (state level) and Select and render UPS Facilities GEO
Deselect (un render) UPS facilities
Some of the facilities remain

Even if you render without zooming out, then deselect the layer, if you zoom out and move you will see some left over facilities in areas you were never over

Ditto with Prison boundaries, in fact seems to be a problem with every GEO dataset
Seems to be only 1 each time

drg101:
> Bug-
> Zoom out (state level) and Select and render UPS Facilities GEO
> Deselect (un render) UPS facilities
> Some of the facilities remain
> 
> Even if you render without zooming out, then deselect the layer, if you zoom out and move you will see some left over facilities in areas you were never over
> 
> Ditto with Prison boundaries, in fact seems to be a problem with every GEO dataset
> Seems to be only 1 each time

@Marc-9 Fixed

drg101:
> Shape length property still leaking through on the Trails set.
> image
> 
> The major problem is fixed, though.
> 
> One thing to note for a future PR is that unchecking some of the really point-dense datasets can take a long time, so it may be worth looking for optimizations here.

@pierce-smith1 Fixed

PR #382, opened 2021-08-12
Fix unresponsive hovering on covid graph

pierce-smith1:
This _should_ address county hovering and highlighting not working - only tested on FF, so make sure it works on your browser first.


Marc-9 (approved):
It works in safari but I could not test in Firefox because of the menu issue
Which brings up the point if the menu does not load and you open the covid graph you are bombarded with errors, this should not be an issue however as if the menu is failing then we should not expect anything to work

PR #383, opened 2021-08-13
Fix newly created histogram bars being black

pierce-smith1:
Should address the issue of histograms sometimes having black bars. Only tested on FF, etc. etc.


Marc-9 (approved):
I cant test because of the menu error, but the code makes sense thanks for this Pierce!

PR #384, opened 2021-08-19
Style popout

honeybadgerofdoom:
There are a couple console errors I have noticed in certain situations than are unrelated to the styling. These bugs don't interfere with functionality but should be addressed in another PR.


drg101 (changes requested):
Everything logically works very smoothly, but could we put the dropdown caret on the far right side of the entry, aswell as only allocate it a small amount of horizontal space?

pierce-smith1 (commented):
Works well and looks very nice. There are potentially two changes here that would improve the UX:
* have an indicator for which field is being used for coloring that's visible outside the dropdown
* change the "color me" checkbox to a radio button, since a checkbox implies any combination of options is possible which is not the case

Also, is it just me not understanding how the temporal dropdown works or the implementation being missing, or does it not do anything? Looking at Covid county and heatwave data, I can't see any changes anywhere when I select different options in the dropdown.

@honeybadgerofdoom I'll approve or request changes once you give me your thoughts on these ideas.

honeybadgerofdoom:
@drg101 @pierce-smith1 love the suggestions, I'll work those ideas in. Pierce, the temporal dropdown isn't connected to anything yet, that will come in a future PR.

honeybadgerofdoom:
@drg101 @pierce-smith1 I moved the dropdown icon right and switched to radios. I didn't add the current color field visibility outside of the dropdown yet because I couldn't find a way to do that without hating the way it looked (lol), but it would be pretty easy to add later on if need be.

drg101 (approved):
Looks good to me, we can talk at the next group meeting about how to indicate that a field is the current color-code field.

PR #392, opened 2021-08-29
Save and load workspace

drg101:
The buttons on top of the menu are now functional. This implementation uses local storage, but the seriliazilation/de-serialization is simple and just outputs a string, so this will be easy to scale-up.


drg101:
You should test this will with date, sliders, selectors, etc. The system stores:

- Current layers in workspace
- State of the layers in workspace (on/off)
- State of constraints in each layer (on/off) AND the state of the slider/checkboxes/date-selector in each.

This means that saving the menu should save ALL of the current state of the menu. Please test this well.

pierce-smith1 (approved):
Genuinely cannot find anything wrong with this. The exact set of datasets is preserved, all constraint parameters are conserved, it correctly remembers which ones are active, date selectors work fine, it really Just Works:tm:. Very well done.

We could do with a little more UX work (e.g. there should probably be some feedback saving was successful), but that is for another PR.

honeybadgerofdoom (approved):
This seems to work exactly as it should. 

A couple things to consider, possibly for future PR's:
1) Save the state of each Layer open/close (whether or not it is dropped down in the menu)
2) Save the state of the Intersections On/Off switch, so users can get back exactly to where they were during a certain session.

And of course I agree with Peirce that we should provide feedback on successful save/load

Marc-9:
Looking good so far however I believe we should also store whether or not we have intersections turned on or not

Marc-9 (approved):

Marc-9:
Once I have finished the session logging and dashboard I believe I will work with Matt to use Daniels implementation here of the serialization to store the users workspace in the database.
Ideally some of the functionality will be
1. Saving multiple workspaces
2. Naming workspace
3. Warning user that when they load a workspace it will overwrite what they currently have stored in their workspace

drg101:
@Marc-9 @pierce-smith1 @pierce-smith1 Thanks for the feedback, I will add the intersection to state in the next PR. The UI for naming and saving will be planned over the next week or so.

PR #393, opened 2021-08-30
Style charting

honeybadgerofdoom:
I'm working on the merge conflict right now. This isn't totally done but this is the first 80% of styling. Doing this now so the branch isn't out for very long, will do another styling PR soon.


honeybadgerofdoom:
Looking at all of the console.log() error with Pierce, coming from Dependency issue

honeybadgerofdoom:
Styling is not done yet, want to get this branch in now so its not out for a super long time. Will finish up styling in next PR.

src/js/ui/NewCharting/ChartGlobalControls.js
```diff
@@ -56,32 +56,69 @@ You may add Your own copyright statement to Your modifications and may provide a
...
+                <Button variant="outlined" startIcon={<ExpandMoreIcon/>} onClick={handleClick}>
+                    Select Chart Type
+                </Button>
+                <Menu
+                    anchorEl={anchorEl}
+                    keepMounted
+                    open={Boolean(anchorEl)}
+                    onClose={handleClose}
+                >
+                    <MenuItem onClick={() => {
```
drg101, line 102: This should be refactored, but this isn't really your responsibility. @Marc-9 @pierce-smith1 Think about making this based on some external object and using a map in a future PR.

drg101 (approved):
Looks very clean and is pretty easy to use. The switch for county/tracts is a bit confusing but that can be easily changed whenever.

Marc-9 (changes requested):
The delete button disappears if the resizable is made to small, consider allowing the menu flow into the second line or increasing the minimum resizable width
Mac safari throws error in console on load
Warning: React does not recognize the `justifyContent` prop on a DOM element

The county and tract toggle button no longer works on correlogram and it seems to always default to county

Marc-9 (approved):
Thanks looks good

src/js/ui/NewCharting/ChartGlobalControls.js
honeybadgerofdoom: @drg101 did a quick refactor here

PR #396, opened 2021-09-01
Workspace from URL

drg101:
Allows the workspace to be seeded by the URL.

To test, you can add this string to your URL in the browser and reload:
`?workspace=N4IgNghgngpgTgZxALgNqgMYHsxhhgFwEssA7FEAuCQgfQFsYATIiU2gCywFcEYuwTWkVLZGIADQgyKKtxhSYADwAObJswDKBCARgoAZhDB8p2UgioQRBbbv1oQAJgAMARhcNmrdl178cIRExfQBdMzJLahskNFCAXwlMHDxCEnJkEGxuUgIoLxY2Th4+ASDRLHEpGWQ5BRBlNVINJjs9WTh5CIsrGzaHVGd3T0ZC3xKAwWEK8XCsyN7c2PQQUghxTNcPAp9i-zLpkMkQS3sUVCcAFgBOa4kAVgA2W+uEhKT53HxiGpOANyItGyuXyAHEAJKaABSAHlwQA5Y41OqKVTqLQ6drIIwmermKLWXL9c4gABKAAUADK0AAqAAkAKIAWQZmhAc3xiwIy3eyS+aV+CABtCsdAh0LhiOqGRRDTRzQxZ1qnTxC2iRMxAzJVNpjJZbI5asJ3POoDWG211PpzNZx1OWNQLgkLgAdABme5veJzGzwPiEDryeJAA`

This should seed your menu with this (it may take awhile to load because it has to wait for all the options in the menu to be loaded before it loads in): image

### DIFF
--- a/src/js/ui/NewDataExploration/Workspace.js
+++ b/src/js/ui/NewDataExploration/Workspace.js
@@ -144,7 +144,8 @@ export default React.memo(function Workspace() {
         setWorkspace(layers.map(layer => {
             const isIn = collections.has(layer.collection);
             if(isIn) {
-                const deSerializedLayer = deSerializedWorkspace.layers.find(e => e.collection === layer.collection);                layer.on = deSerializedLayer.on;
+                const deSerializedLayer = deSerializedWorkspace.layers.find(e => e.collection === layer.collection);                
+                layer.on = deSerializedLayer.on;
                 layer.expandedState = deSerializedLayer.expandedState;
                 layer.constraintState = deSerializedLayer.constraintState;
                 layer.forceUpdateFlag = true;
--- a/src/js/ui/NewDataExploration/Workspace.js
+++ b/src/js/ui/NewDataExploration/Workspace.js
@@ -56,7 +56,7 @@ You may add Your own copyright statement to Your modifications and may provide a
 
 END OF TERMS AND CONDITIONS
 */
-import React, {useEffect, useState} from 'react';
+import React, {useEffect, useRef, useState} from 'react';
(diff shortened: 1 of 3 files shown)


drg101:
@Marc-9 Could you set it up so the "workspace" URL param gets forwarded on the live site to the iFrame. This will allow us to seed aperture with a specific workspace on the live-site by modifying the `workspace` parameter of: `https://urban-sustain.org/services/aperture.php?workspace=<workspaceString>`.

Marc-9:
> @Marc-9 Could you set it up so the "workspace" URL param gets forwarded on the live site to the iFrame. This will allow us to seed aperture with a specific workspace on the live-site by modifying the `workspace` parameter of: `https://urban-sustain.org/services/aperture.php?workspace=<workspaceString>`.

I can have this done in 3-5 business days, or you can request express shipment which will cost you an emoji

honeybadgerofdoom (approved):
Well this is super cool

src/js/ui/NewDataExploration/Workspace.js
```diff
@@ -77,17 +77,29 @@ export function prettifyJSON(name) {
     return Util.capitalizeString(Util.underScoreToSpace(name));
 }
 
+let workspaceIsLoaded = false;
```
pierce-smith1, line 80: Having globals float around feels like bad practice, is there a better way to do this?

drg101: Yes, in state, I'll change it to this

pierce-smith1 (approved):
React continues to make less and less sense by the day...

PR #397, opened 2021-09-01
Add NSF logo, with tooltip text, to lower-right of map

honeybadgerofdoom:
The idea here being that we can remove that footer from the /aperture page on the website, because it is problematic. TBH this actually looks a lot better than I thought it would.

Putting this up as a draft to see what y'all think, will confer with Shrideep Friday and see what he thinks.


drg101:
I like this, tooltip text could be made bigger though. This makes aperture much more "portable," IE we (or anyone) can put the iFrame on any website and it still has the NSF blurb and such.

honeybadgerofdoom:
> I like this, tooltip text could be made bigger though. This makes aperture much more "portable," IE we (or anyone) can put the iFrame on any website and it still has the NSF blurb and such.

That's a really good point.

pierce-smith1:
Yeah, I'm a fan of how this looks and feels. Hopefully it can get approved.

honeybadgerofdoom:
Shrideep approves

PR #398, opened 2021-09-02
Fix workspace performance

drg101:
Fixes some gripes I had with the performance of the workspace.

- Memoized some things so adding a lot of layers isn't progressively slower as you add more
- Only gets whats "needed" from the server on startup, meaning we don't have to wait as long for the menu to load.


pierce-smith1 (approved):
Looks like everything still works.

PR #400, opened 2021-09-03
Import the image instead of finding path in html

honeybadgerofdoom:
No certain if this will fix it, but I think it may. Unfortunately we won't actually know until we merge it


pierce-smith1 (approved):
Well, at the very least it doesn't break my local version.

drg101 (approved):

drg101:
Yea this should work.

PR #404, opened 2021-09-06
Style the Correlogram to match other graphs.

honeybadgerofdoom:


pierce-smith1 (approved):
Looks good and doesn't break anything.
The correlogram is far from perfect of course - eventually we need to address these labels being too long and potentially change the text color/size so it's easier to read.
image

drg101 (approved):

PR #406, opened 2021-09-08
Saving & Loading Improvements + Share Button & More

drg101:


drg101:
Yes I know the styling is bad, @honeybadgerofdoom Plz fix the styling on `Save.js`, `Load.js`, `Share.js`, `SaveAndLoadAndShare.js`, and `SavedWorkspaceSlotSelection.js` once this is merged.

- Added menus for saving, loading, & sharing
- There is 10 slots to save with
- Color state gets saved now. **Known bug: The saved/loaded color of intersected tracts/counties is inconsistent**

The live site (which is what the share link defaults to), doesn't support view port or color state right now, so you will just have to trust me on that for now OR add the `?workspace=<base64String>` into your localhost environment.

drg101:
Please test the _functionality_ of this as well as you can. I threw this together pretty quick.

src/js/ui/NewDataExploration/Save.js
```diff
@@ -0,0 +1,121 @@
...
+    const [name, setName] = useState(`Saved Workspace`)
+    const validName = name.length !== 0;
+
+    const saveWorkspace = () => {
+        const serializedWorkspace = serializeWorkspace(name, saveColor, saveViewport);
+        localStorage.setItem(`workspace${slotCurrentlySelected}`, serializedWorkspace)
+        setModalOpen(false)
+    }
+
+    if (componentIsRendering) { console.log("|Save Rerending|") }
```
pierce-smith1, line 91: Rogue log

pierce-smith1, line 91: Rogue log

src/js/ui/NewDataExploration/SaveAndLoadAndShare.js
```diff
@@ -0,0 +1,114 @@
...
+        boxShadow: theme.shadows[5],
+        padding: theme.spacing(2, 4, 3),
+    }
+}));
+
+
+export default React.memo(function SaveAndLoad({ mode, modalOpen, setModalOpen, serializeWorkspace, deSerializeWorkspace }) {
+    const classes = useStyles();
+
+    if (componentIsRendering) { console.log("|SaveAndLoad Rerending|") }
```
pierce-smith1, line 88: Ditto

src/js/ui/NewDataExploration/SavedWorkspaceSlotSelection.js
```diff
@@ -0,0 +1,122 @@
...
+
+
+export default React.memo(function SavedWorkspaceSlotSelection({slotCurrentlySelected, setSlotCurrentlySelected, onlyShowFullSlots}) {
+    const classes = useStyles();
+
+    const getWorkspace = (index) => {
+        return localStorage.getItem(`workspace${index}`)
+    }
+
+    if (componentIsRendering) { console.log("|Load Rerending|") }
```
pierce-smith1, line 84: Ditto

src/js/ui/NewDataExploration/Load.js
```diff
@@ -0,0 +1,96 @@
...
+    const classes = useStyles();
+
+    const [slotCurrentlySelected, setSlotCurrentlySelected] = useState(1)
+
+    const loadWorkspace = () => {
+        deSerializeWorkspace(localStorage.getItem(`workspace${slotCurrentlySelected}`))
+        setModalOpen(false)
+    }
+
+    if (componentIsRendering) { console.log("|Load Rerending|") }
```
pierce-smith1, line 86: Ditto

pierce-smith1 (changes requested):
I tried a lot of different things and it all seems to be working (nearly) flawlessly. Small issues:

* You can save a workspace with an empty name (it still works, but you probably shouldn't be able to do this)
* There are 9 slots instead of 10
* Several logs left behind in code

From a design standpoint, I can forsee some issues with saving workspaces in local storage. I don't know how permanent it is, but if there's a potential that it disappears without warning that could be a problem. Saving as files _may_ be a better alternative, depending on how reliable local storage is.

drg101:
> I tried a lot of different things and it all seems to be working (nearly) flawlessly. Small issues:
> 
> * You can save a workspace with an empty name (it still works, but you probably shouldn't be able to do this)
> * There are 9 slots instead of 10
> * Several logs left behind in code
> 
> From a design standpoint, I can forsee some issues with saving workspaces in local storage. I don't know how permanent it is, but if there's a potential that it disappears without warning that could be a problem. Saving as files _may_ be a better alternative, depending on how reliable local storage is. (Admittedly, local storage is much easier and more intuitive to use than files)

- These logs don't show up during normal use, they are in all the workspace react files for when we want to see if they are rendering. The flag can be turned on in an external file.

- I'll add one more slot, thanks for pointing that out.

- Local storage should be fine for now, as it is only deleted if the [user wants it to be deleted](https://stackoverflow.com/questions/8537112/when-is-localstorage-cleared). Anyways, we will probably be transitioning this to using the API sometime in the future.

pierce-smith1 (approved):

Marc-9:
Would you say the save workspace overlay would be under the purview of styling?
Atm on my Mac I cant save to anything other than the first 5 slots as the others are hidden beyond my screen, most likely would be solved by changing the default coords or doing an overflow:hidden
In fact I can not even save because it is hidden

honeybadgerofdoom (approved):
This is super duper cool. Will style it once it's merged.

Marc-9 (approved):
I don't know enough about local storage to know if there are some reserved names, but we may want to sanitize/limit the chars allowed for saving a name
Also perhaps a different button/disable the load workspace if no workspaces are saved

pierce-smith1:
> I don't know enough about local storage to know if there are some reserved names, but we may want to sanitize/limit the chars allowed for saving a name

This shouldn't be a problem, since the workspace names are stored in the serialized JSON rather than keys in local storage. I tried a thousand-character long name and one with unicode characters and both worked.

drg101:
> I don't know enough about local storage to know if there are some reserved names, but we may want to sanitize/limit the chars allowed for saving a name
> Also perhaps a different button/disable the load workspace if no workspaces are saved

Luckily the 'name' is contained within the serialized & compressed workspace data, and the workspaces are named `workspace<i>` where `i` is a number from 1 to 10.

drg101:
> > I don't know enough about local storage to know if there are some reserved names, but we may want to sanitize/limit the chars allowed for saving a name
> 
> This shouldn't be a problem, since the workspace names are stored in the serialized JSON rather than keys in local storage. I tried a thousand-character long name and one with unicode characters and both worked.

+1

drg101:
> Also perhaps a different button/disable the load workspace if no workspaces are saved

Good idea

PR #411, opened 2021-09-09
Style Save/Load/Share

honeybadgerofdoom:
This PR styles the save/load/share components. Open to suggestions as always.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,10677 @@
 {
   "name": "aperture",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "aperture",
+      "version": "1.0.0",
+      "license": "ISC",
+      "dependencies": {
+        "@date-io/moment": "^1.3.13",
+        "@material-ui/core": "^4.11.4",
+        "@material-ui/icons": "^4.11.2",
+        "@material-ui/lab": "^4.0.0-alpha.58",
+        "@material-ui/pickers": "^3.3.10",
+        "@nivo/core": "^0.73.0",
+        "@nivo/heatmap": "^0.73.0",
+        "@nivo/line": "^0.73.0",
+        "@popperjs/core": "^2.9.3",
+        "bootstrap": "^5.1.0",
+        "calculate-correlation": "^1.2.2",
+        "canvas": "^2.6.1",
+        "clsx": "^1.1.1",
+        "dexie": "^3.0.3",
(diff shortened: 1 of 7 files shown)


pierce-smith1 (changes requested):
Needless to say this is a _massive_ improvement. Some suggestions/comments:

Workspace names should be truncated if they are too long, otherwise it bleeds into the radio button and causes a nasty horizontal scroll to appear.
image

This is a nice indication of what should go into this box, but it also means that if they don't put anything down, the workspace gets saved as "Workspace Name..." rather than just being rejected.
image
On this topic as well: when a workspace save gets rejected due to a missing name, there should be some indication that is why the saving failed - right now, absolutely nothing happens.

This one could go either way. Personally I'm not a fan of the way that clicking in the text portion of the box automatically copies the whole link - what if I just want to copy part of the link? That being said, 99% of the time people just want to copy the whole thing and the current behavior makes that a lot easier. (I'm sure there's a good hour or two of possible UX discussion here so I'm happy to hear other thoughts)
image

High screen DPIs/zoom levels cause this to get skewed off-center a bit - it may be possible to get better centering.
image

Marc-9 (approved):
Nothing new from what Pierce said, I would like however to be able to select a workspace by clicking its row and not just by clicking the button.
Not necessary at all, I have no idea how easy/hard this would be

drg101:
> Needless to say this is a _massive_ improvement. Some suggestions/comments:
> 
> Workspace names should be truncated if they are too long, otherwise it bleeds into the radio button and causes a nasty horizontal scroll to appear.
> image
> 
> This is a nice indication of what should go into this box, but it also means that if they don't put anything down, the workspace gets saved as "Workspace Name..." rather than just being rejected.
> image
> On this topic as well: when a workspace save gets rejected due to a missing name, there should be some indication that is why the saving failed - right now, absolutely nothing happens.
> 
> This one could go either way. Personally I'm not a fan of the way that clicking in the text portion of the box automatically copies the whole link - what if I just want to copy part of the link? That being said, 99% of the time people just want to copy the whole thing and the current behavior makes that a lot easier. (I'm sure there's a good hour or two of possible UX discussion here so I'm happy to hear other thoughts)
> image
> 
> High screen DPIs/zoom levels cause this to get skewed off-center a bit - it may be possible to get better centering.
> image

Thanks for doing this quick. I'm just gonna merge this and implement these bug fixes in my next PR since they are pretty minor.
@pierce-smith1 I think the whole text getting copied is fine since the only other thing people would want is the base URL, which will already be the URL in their browser.

drg101 (approved):

PR #414, opened 2021-09-10
General Polishing and Improvements

drg101:
- Fix table values that made no sense like epoch time not showing up as a date.
- Limit floats in the table to 3 decimal places
- Temporarily disable modeling tab
- Range constraints go grey when they don't actually affect the query
- Risk index constraints now have sliders again
- A thing tells you if a query is happening
- Fix the little bugs in save & load, limiting single word length
- Menu title on popup makes more sense now
- Units show up again on popup

Sadly wasn't able to fix the intersection color bug yet, but luckily no one should notice as it doesn't affect interactability.

### DIFF
--- a/src/css/main.css
+++ b/src/css/main.css
@@ -173,11 +173,11 @@ body {
 .warningContainer{
   position: absolute;
   z-index: 1999;
-  bottom: 0;
+  bottom: 6vw;
   right: 0;
   width: 0;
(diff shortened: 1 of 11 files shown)


honeybadgerofdoom (approved):
Nothing breaks on my end - there's a lot in this PR so it's hard to test everything

PR #417, opened 2021-09-13
Small polish to COVID chart

pierce-smith1:
Adds labels to the axes and removes the query notice (the "getting data..." text).


drg101 (approved):
Looks good.

PR #418, opened 2021-09-13
Fix Bug Relating to Zooming and Changing Slider

drg101:
Basically, a fix I made earlier made it so the query class could return immediately. The query id set was not populated yet once it returned, causing many issues. This populates the id set before the call is made.


pierce-smith1 (approved):
LGTM

Marc-9 (approved):
Ah yes the classic move the code up 9 lines trick, a favorite of mine

PR #419, opened 2021-09-15
Remove 'graph me' and 'reset constraints' from layer controls (for now)

drg101:
Refer to title.


Marc-9 (approved):
may cause breaking changes

PR #420, opened 2021-09-15
Add Safe Query Logging

drg101:
The request is is in a try-catch loop so it shouldn't break anything.


drg101:
Trying to remove the console error logging.

drg101:
@Marc-9 This now reflects the ones I was sending you

Marc-9 (approved):
Thanks for getting this done Daniel, looking forward to seeing the results of this
(every time the map is moved this issues a new query for each layer correct?)
should we monitor the date time the query was made at?

PR #422, opened 2021-09-16
Confirm overwrite workspace

honeybadgerofdoom:
This PR adds an alert asking users to confirm before they overwrite an existing workspace

### DIFF
--- a/src/js/ui/NewDataExploration/Load.js
+++ b/src/js/ui/NewDataExploration/Load.js
@@ -56,6 +56,64 @@ You may add Your own copyright statement to Your modifications and may provide a
 
 END OF TERMS AND CONDITIONS
 */
+/*                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+
+Software in the Sustain Ecosystem are Released Under Terms of Apache Software License
+
+This research has been supported by funding from the US National Science Foundation's CSSI program through awards 1931363, 1931324, 1931335, and 1931283. The project is a joint effort involving Colorado State University, Arizona State University, the University of California-Irvine, and the University of Maryland - Baltimore County. All redistributions of the software must also include this information.
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+
+1. Definitions.
+
+"License" shall mean the terms and conditions for use, reproduction, and distribution as defined by Sections 1 through 9 of this document.
+
+"Licensor" shall mean the copyright owner or entity authorized by the copyright owner that is granting the License.
+
+"Legal Entity" shall mean the union of the acting entity and all other entities that control, are controlled by, or are under common control with that entity. For the purposes of this definition, "control" means (i) the power, direct or indirect, to cause the direction or management of such entity, whether by contract or otherwise, or (ii) ownership of fifty percent (50%) or more of the outstanding shares, or (iii) beneficial ownership of such entity.
+
+"You" (or "Your") shall mean an individual or Legal Entity exercising permissions granted by this License.
+
+"Source" form shall mean the preferred form for making modifications, including but not limited to software source code, documentation source, and configuration files.
+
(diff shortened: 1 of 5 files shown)


pierce-smith1 (commented):
Works:tm: beautifully, will approve once the CC issues are fixed. You can probably ignore the cognitive complexity one but the code duplication issues seem valid.
(On that note, why are we just now getting those two code duplication issues? Hasn't that code existed for a while)?

src/js/ui/NewDataExploration/Load.js
```diff
@@ -54,6 +54,64 @@ You may add Your own copyright statement to Your modifications and may provide a
 
 9. Accepting Warranty or Additional Liability. While redistributing the Work or Derivative Works thereof, You may choose to offer, and charge a fee for, acceptance of support, warranty, indemnity, or other liability obligations and/or rights consistent with this License. However, in accepting such obligations, You may act only on Your own behalf and on Your sole responsibility, not on behalf of any other Contributor, and only if You agree to indemnify, defend, and hold each Contributor harmless for any liability incurred by, or claims asserted against, such Contributor by reason of your accepting any such warranty or additional liability. 
 
+END OF TERMS AND CONDITIONS
```
drg101, line 57: Extra licence was inserted here for some reason.

src/js/ui/NewDataExploration/Load.js
```diff
@@ -101,7 +159,7 @@ export default React.memo(function Load({deSerializeWorkspace, setModalOpen}) {
                 <Typography className={classes.title} align="center" variant="h5">Load Workspace</Typography>
             </Grid>
             <Grid item className={classes.fullWidth}>
-                <SavedWorkspaceSlotSelection title="Select a Saved Workspace" slotCurrentlySelected={slotCurrentlySelected} setSlotCurrentlySelected={setSlotCurrentlySelected} onlyShowFullSlots/>
+                <SavedWorkspaceSlotSelection title="Select a Saved Workspace" setNameOfCurrentSlot={() => {}} slotCurrentlySelected={slotCurrentlySelected} setSlotCurrentlySelected={setSlotCurrentlySelected} onlyShowFullSlots/>
```
drg101, line 162: Why is setNameOfCurrentSlot just a empty function, should this be deleted?

drg101 (changes requested):
A few things on the logic of how it is determined that a slot already has a name.
Right now, the `setNameOfCurrentSlot` is used to pass this up to the `Save` component. The `Save` component already has a reference to the current slot with `slotCurrentlySelected` though, so this other prop is redundant. 

Looks good other than this though, will approve once the state stuff is addressed.

src/js/ui/NewDataExploration/Load.js
honeybadgerofdoom: It needs to be there or SavedWorkspaceSlotSelection throws an error when called from Load.js.

honeybadgerofdoom:
> A few things on the logic of how it is determined that a slot already has a name.
> Right now, the `setNameOfCurrentSlot` is used to pass this up to the `Save` component. The `Save` component already has a reference to the current slot with `slotCurrentlySelected` though, so this other prop is redundant.
> 
> Looks good other than this though, will approve once the state stuff is addressed.

slotCurrentlySelected is an int. If I understand correctly it is used to index into a data structure that isn't visible in Save.js, so I created another piece of state to hold the name (or null) at that index. Not sure how I could use just this int to get the name of the selected slot, or if it is null.

honeybadgerofdoom:
> A few things on the logic of how it is determined that a slot already has a name.
> Right now, the `setNameOfCurrentSlot` is used to pass this up to the `Save` component. The `Save` component already has a reference to the current slot with `slotCurrentlySelected` though, so this other prop is redundant.
> 
> Looks good other than this though, will approve once the state stuff is addressed.

K I'll work on this, thanks!

drg101 (changes requested):
One more thing, could "Empty Slot" be changed to something that makes more sense like "Saved Workspace" or something?

PR #423, opened 2021-09-17
Alerts user on save/load/delete success

honeybadgerofdoom:


drg101:
@honeybadgerofdoom I changed one naming thing, but everything else looks good.

drg101 (approved):

Marc-9 (approved):
Alls looks good to me, perhaps we don't need a save and a load and just one manage workspace button, I as a super smart user know that I can delete a workspace by clicking save then clicking an existing one and pressing delete, but perhaps that is not obvious

PR #424, opened 2021-09-19
No more nav tabs

honeybadgerofdoom:
- Removed navigation tabs as this is now a data-visualizing only app
- Most of the Files Changed are switching the {componentIsRerendering} import to ../Sidebar.js, because TabSystem.js has been removed.


pierce-smith1 (approved):
LGTM
There is one very minor issue here with the Intersections button: if you toggle it on and off the change in text length causes the other buttons to shift around slightly, which is a bit strange. I'll leave fixing that up to you since it's so minor.

PR #425, opened 2021-09-21
More info link to website

honeybadgerofdoom:
This is a proof-of-concept PR that Shrideep had asked for a while back. It creates a button for each Layer that links to the website where the Layer's dataset info is. We should add this URL to the metadata catalogue so each Layer has that field and we can access it via props.layer.datasetInfoPath.


honeybadgerofdoom:
I'll fix these conflict tomorrow.

PR #426, opened 2021-09-24
organize workspace buttons into groups

honeybadgerofdoom:
Tiny PR to organize Workspace Control buttons into groups - a little less clutter


pierce-smith1 (changes requested):
Works:tm:, but I will admit it's a little strange that the intersections, graph, and close buttons are grouped since there's no obvious relationship between them. Maybe this should only apply to save/load/share?

drg101 (commented):
+1 For what @pierce-smith1 said, maybe we should re-think this row entirely since it was just ported over from the version with the extra tabs.

honeybadgerofdoom:
I agree - it's pretty fine as is but could definitely be better. I think re-considering the whole workspace control section is in order.

PR #432, opened 2021-11-08
Log Druid queries

pierce-smith1:
Logging to the query API is now a standalone function and Druid queries do it too. @Marc-9 Make sure the payload object looks okay, I changed it a little bit.

### DIFF
--- a/src/js/library/Query.js
+++ b/src/js/library/Query.js
@@ -106,7 +106,7 @@ const Query = {
     },
 
     /**
-      * Posts the given query to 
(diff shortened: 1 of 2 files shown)


drg101 (approved):
Code looks good, if @Marc-9 says the data is logging then all is good.

Marc-9:
> Code looks good, if @Marc-9 says the data is logging then all is good.

The original mongo queries were logging, unable to test new druid queries until that service is up and running

Also the random - all queries being logged under my account - seems to have disappeared??
Weird

PR #433, opened 2021-11-10
Refactor NewDataExploration file structure, change dataset searchbar

honeybadgerofdoom:
This PR looks huge because it moves all the files in NewDataExploration around, but it only actually makes meaningful changes to WorkspaceSearchbar.js

- I've been wanting to change this around for a while. The issue is that previously, when you searched for a dataset and selected it, the search would automatically reset. Thus, if you wanted to add all County datasets, you would have to type 'county' over and over again, clicking one at a time. This PR not only fixes that, but adds an 'add all' button when filtering, so you can add all of your search results to the workspace at once.
- Put the results of the search in a table, which I think looks a bit nicer.
- Fixes a couple .maps.

This is a draft for now, let me know what you all think/any ideas for improvements.


honeybadgerofdoom:
Ugh I checked this out of an older version of master. Will remake

PR #434, opened 2021-11-10
New dataset searchbar

honeybadgerofdoom:
This PR fixes some issues I had with the old search bar.

- I've been wanting to change this around for a while. The issue is that previously, when you searched for a dataset and selected it, the search would automatically reset. Thus, if you wanted to add all County datasets, you would have to type 'county' over and over again, clicking one at a time. This PR not only fixes that, but adds an 'add all' button when filtering, so you can add all of your search results to the workspace at once.
- Put the results of the search in a table, which I think looks a bit nicer.
- To test this, just play around with searching, adding, and removing datasets to/from the workspace.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1878,7 +1878,7 @@
     "@xtuc/ieee754": {
       "version": "1.2.0",
       "resolved": "https://registry.npmjs.org/@xtuc/ieee754/-/ieee754-1.2.0.tgz",
-      "integrity": "sha512-DX8nKgqcGwsc0eJSqYt5lwP4DH5FlHnmuWWBRy7X0NcaGR0ZtuyeESgMwTYVEtxmsNGY+qit4QYT/MIYTOTPeA==",
+      "integrity": "sha1-7vAUoxRa5Hehy8AM0eVSM23Ot5A=",
       "dev": true
     },
     "@xtuc/long": {
@@ -1895,7 +1895,7 @@
     "abbrev": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
-      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q=="
+      "integrity": "sha1-+PLIh60Qv2f2NPAFtph/7TF5qsg="
     },
     "accepts": {
       "version": "1.3.7",
@@ -1997,7 +1997,7 @@
     "anymatch": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/anymatch/-/anymatch-2.0.0.tgz",
-      "integrity": "sha512-5teOsQWABXHHBFP9y3skS5P3d/WfWXpv3FUpy+LorMrNYaT9pI4oLMQX7jzQ2KklNpGpWHzdCXTDT2Y3XGlZBw==",
+      "integrity": "sha1-vLJLTzeTTZqnrBe0ra+J58du8us=",
       "dev": true,
       "requires": {
         "micromatch": "^3.1.4",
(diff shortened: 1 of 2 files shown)


pierce-smith1 (changes requested):
I agree this looks a lot nicer, though I find the layout a bit misleading:

image
* The "Clear All" and "Add All" buttons don't look like buttons at all
* It's not immediately clear what "Add/Clear All" do (does "Add All" add _every_ dataset or just the ones in the search results? does "Clear All" clear _every_ enabled dataset or just the ones in the search results? I know the answer, rhetorical question)
* The way the "Clear All", "X matching datasets", and "Add All" texts are aligned with the table columns makes them look like the names of the columns, but of course they have no such relation
* Since the graphing and info tooltip icons aren't actually buttons, they shouldn't act like buttons (reacting to clicks, becoming a shaded circle when hovered over, etc.)

honeybadgerofdoom:
@pierce-smith1 great feedback, thanks! I moved the 'add all' button to the dataset column, as I believe there is connection between adding all and the datasets themselves. Further, when you are filtering, the Add All <> To Workspace appears as the blue button color.

For clearing all, I think there is a relation between 'Clear Workspace' and the checkbox column, as this is the column you interact with in order to add/subtract things to the workspace, it makes sense that you'd subtract all from there as well. This button now conditionally renders when there are datasets in the workspace, and appears as red.

Let me know what you think. The other thing that makes sense to me conceptually is putting both buttons in that column so that all dataset adding/removing happens in one place. It's just visually difficult to achieve this.

pierce-smith1:
The button labels are extremely clear now, very nice!

> The other thing that makes sense to me conceptually is putting both buttons in that column so that all dataset adding/removing happens in one place. It's just visually difficult to achieve this.

I agree, if the buttons belong in _any_ column, they belong in the checkbox column, since both operate _on_ that column. Maybe you could just have a single button in the checkbox column that toggles all visible checkboxes. That allows you to have add and clear all with little visual clutter ("Clear Workspace" might still need to be a separate button).

I like the idea of using icons for these buttons in general since if they are just text living in the header row, like they are right now, they look to me like _labels_ trying to tell me _about_ the table, not actions I can take on it. Icons with tooltips would get the message across much clearer, I feel.

> As for the graphing and tooltip icons looking like buttons I agree, that's not ideal. That is left over from the old UI and should be addressed - I can tackle that in another PR

Yeah, my comment on that was off topic to this PR. Glad you agree though.

honeybadgerofdoom:
@pierce-smith1 Have a look - I conditionally render both as icon buttons with tooltips in the checkbox column header.

pierce-smith1 (approved):
Sorry for taking a whole day to review this - I would have done so sooner but apparently this update got lost in a flurry of github emails.

I really love this change. I think there are a few smaller UX things to consider (should the minus clear the whole workspace or just the ones matching search?) but in its current state this is awesome and I'm happy to approve it.

drg101 (changes requested):
Looks very pretty and I like the UI.
~~I'm noticing alot of flickering when opening and using the menu though.~~
~~Here is a link: https://streamable.com/vmdafk~~

On a less important note, could it say "x results for <search>", for example in this situation it would say "Found 14 Results for 'Neon'"

Screenshot 2021-11-11 5 08 21 PM

pierce-smith1:
> I'm noticing alot of flickering when opening and using the menu though.
Here is a link: https://streamable.com/vmdafk

On Linux/Firefox I am not experiencing this.

honeybadgerofdoom:
@pierce-smith1 I like the text update idea - I made that change. Also not noticing the flickering on my end (using mac/Chrome)

honeybadgerofdoom:
> @drg101 I like the text update idea - I made that change. Also not noticing the flickering on my end (using mac/Chrome)

pierce-smith1:
RIP me, accidentally committed to this branch instead of mine, please disregard...

drg101 (changes requested):
Looks very good, the only thing I'd like to see now is for the the list to expand when the searchbar textbox is clicked (Right now you have to click the dropdown arrow to see this). I think it just makes more sense to give the user a list of datasets the moment they click on the searchbar to type something in.

honeybadgerofdoom:
@drg101 good idea, I added that.

drg101 (approved):
Looks good.

PR #435, opened 2021-11-10
Fix a couple sloppy iterative renders

honeybadgerofdoom:
This fixes 2 files where I had (stoooopidly) pushed DOM elements into an array and rendered that instead of using .map(). I wrote these when I was still getting the hang of things. Don't judge me.

Probably some more aperture refactoring to come - there are a few more little things I want to clean up in the code base.

### DIFF
--- a/src/js/ui/NewDataExploration/Layer.js
+++ b/src/js/ui/NewDataExploration/Layer.js
@@ -87,22 +87,6 @@ function extractLayerConstraints(layer) {
     return [defaultLayerConstraints, allLayerConstraints];
 }
 
-function createConstraints(activeLayerConstraints, allLayerConstraints, classes, querier) {
-    let constraints = [];
-    activeLayerConstraints.forEach((constraint, index) => {
-        if(constraint) {
-            constraints.push(
-                <div key={index}>
-                    <Paper elevation={3}>
-                        <IndividualConstraint constraint={allLayerConstraints[index]} classes={classes} querier={querier} />
-                    </Paper>
-                </div>
-            );
-        }
-    });
-    return constraints;
-}
-
 export default React.memo(function Layer(props) {
     const classes = useStyles();
     const [check, setCheck] = useState(false);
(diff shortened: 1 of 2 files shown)


pierce-smith1 (commented):
I don't think the `forEach`s and `map`s here are exactly equivalent: the old code did not place falsey elements into the rendered lists, but the new one replaces falsey elements with `undefined`. I'm not sure if this causes any problems yet (how does React handle trying to render `undefined`?)

drg101 (approved):
Looks good.

honeybadgerofdoom:
@pierce-smith1 I'm not quite sure I follow. Both .map's are iterating over arrays of boolean values, looking for True. In the event that an entry is True, a component is rendered based off of that index. How would this cause React to try and render _undefined_?

pierce-smith1:
> @pierce-smith1 I'm not quite sure I follow. Both .map's are iterating over arrays of boolean values, looking for True. In the event that an entry is True, a component is rendered based off of that index. How would this cause React to try and render _undefined_?

A map produces an array with exactly one value for every element in the input array. If the map function produces `undefined` for any value in the input array, `undefined` will be put into the new array. In JS, if execution reaches the end of a function with no return statement, `undefined` is returned. The function you're passing into `.map` is roughly
```js
(e) => { if (e) return <Component> }
```
which will return `<Component>` if `e` is truthy or `undefined` if `e` is falsy. The `.map` will use this `undefined` as the entry in the new array.

Try this in a browser console:
```js
[true, false].map(e => { if (e) return 'something' });
```
and you'll see it outputs `['something', undefined]`.

So the maps you wrote will be equivalent to the old for-each loops only if React doesn't do anything when it tries to render undefined, which I think is the case (don't you return `null` out of a component when you want to render nothing?)

pierce-smith1 (approved):
> only if React doesn't do anything when it tries to render undefined

Well, I just did some testing and weirdly, while returning `undefined` directly from a component _does_ cause React to crash, inserting `undefined` into a rendered list like you're doing in this code does not. So... plot twist, everything is actually fine.

That being said I would consider adding a `return null` to the end of the map functions since it seems to be the more canonical way to render nothing.

PR #436, opened 2021-11-12
gridMET time range goes up to yesterday

pierce-smith1:
Lets users select dates up to yesterday for the gridMET dataset, since we will very soon be updating it regularly.

### DIFF
--- a/src/js/static/menuOverwrite.js
+++ b/src/js/static/menuOverwrite.js
@@ -60,6 +60,11 @@ END OF TERMS AND CONDITIONS
 import clone from 'just-clone';
 import Util from '../library/apertureUtil';
 
+function getEpochForYesterday() {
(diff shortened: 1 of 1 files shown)


drg101:
Screenshot 2021-11-12 6 05 07 PM

drg101:
I think this would be a better way to do it, just so that is standardized to the **start** of yesterday, instead of exactly 24 hours ago. It wont affect the functionality but it is more accurate in general.

drg101 (changes requested):
See comment.

pierce-smith1:
> I think this would be a better way to do it, just so that is standardized to the **start** of yesterday, instead of exactly 24 hours ago. It wont affect the functionality but it is more accurate in general.

I agree, this has been changed.

drg101 (approved):
Looks good to me.

PR #438, opened 2021-11-15
refactor file structure

honeybadgerofdoom:
No functional or aesthetic changes here, just re-organizes the files in NewDataExploration into a file structure that makes sense.


drg101 (approved):
I don't think anything has broken from what I tested.

pierce-smith1 (commented):
Everything seems to still work and I like the new structure, though can we come up with something more descriptive than "WorkspaceControls"?

PR #440, opened 2021-12-01
Added Soils Data

drg101:
Added soils data, it may be very slow to query for the next day or so because I still need to set up the indexing on it.

### DIFF
--- a/src/json/menumetadata.json
+++ b/src/json/menumetadata.json
@@ -6843,7 +6843,10 @@
         ]
     },
     {
-        "collection": "soils_statsgo",
+        "collection": "soils_statsgo_clean",
+        "label": "Soils STATSGO",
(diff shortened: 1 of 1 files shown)


pierce-smith1 (changes requested):
Dataset looks really cool! Some strange things:

* It looks like -0.1 is being used as a missing value and it's causing some UI oddities - is there a way you could deal with this without too much trouble?
image
* The name of the dataset isn't super descriptive, plus it could benefit from a help blurb describing what it is.
* This variable is kind of confusing: it sounds like a boolean on/off indicator but it has units?
image

drg101:
> Dataset looks really cool! Some strange things:
> 
> * It looks like -0.1 is being used as a missing value and it's causing some UI oddities - is there a way you could deal with this without too much trouble?
>   image

Sadly not yet, it would cause other issues if I quickly set up the null value to be something else.

drg101:
> * This variable is kind of confusing: it sounds like a boolean on/off indicator but it has units?
>   image

Yea not sure what is going on with this, I didn't download the data but I'll ask Saptashwa.

drg101:
@pierce-smith1 Added some description and fixed 2d indexing, sadly the other things can't really be fixed right now.

pierce-smith1:
Did you forget to push? I don't see any changes to the branch.

honeybadgerofdoom (approved):
Looks pretty good - ditto on @pierce-smith1's notes. Love the way this dataset visualizes with different color fields.

drg101:
> Did you forget to push? I don't see any changes to the branch.

Oops, pushed now.

pierce-smith1 (approved):

PR #441, opened 2021-12-03
Ingest aqi datasets

honeybadgerofdoom:
This adds 6 of the 8 AQI Contaminants. They can be found by searching 'EPA' in the Aperture search bar. Shrideep wants to get these in this morning, if possible, for a PI meeting this afternoon.

Known issue: when a contaminant is added the icon does not automatically color-code - this must be selected in the inspection pane. Shrideep saw this and said it is fine for the time being. 

I will add the last 2 contaminants once they're done being ingested, hopefully by the end of the weekend.

Just check that you can see these 6 datasets, that they turn on, and it doesn't break anything.


pierce-smith1 (approved):
It works - again this could do with more detailed names and descriptions but since this is high priority I will approve for now.

drg101 (approved):
Doesn't break existing stuff.

PR #442, opened 2021-12-06
Change Over Time Query

Marc-9:
This adds a change over time query for all Temporal Datasets.
Testing
Visualize temporal datasets (heatwave, covid, nivo)
Click on a county to open widget, at the bottom several new change over time percentages should be added
You may now set this to be the current color field
Change the years in the menu to weird times and see what happens.
Does the nivo data take too long to load
    -this is because it is sorted, to guarantee the change over time is correct, it should be by default based on they way we insert, but potentially this may change

All divide by 0 errors should be gone
Color Coding will be blank when you set it as the current color field (this will need to be hardcoded in)

Once this is in, development on adding this to the menu as a constraint, where users can constrain on only counties in which the change over time was greater than less than X

EDIT

Please also check EPA dataset still exists along side county change, merge conflict in menu data that I should have resolved properly


drg101:
I think the color chart should be added before we merge this.

Marc-9:
> I think the color chart should be added before we merge this.

Done

src/js/library/autoQuery.js
```diff
@@ -265,12 +265,22 @@ export default class AutoQuery {
         if (fieldName === this.color.variable) {
             predefinedColor = this.color;
         }
-        const colorField = this.data.constraints[fieldName] ?? this.data.constraints[`properties.${fieldName}`] ?? this.data.constraints[fieldName.substring(0,fieldName.indexOf(temporalId))];
+        
+        let colorField = this.data.constraints[fieldName] ?? this.data.constraints[`properties.${fieldName}`] ?? this.data.constraints[fieldName.substring(0, fieldName.indexOf(temporalId))];
+        
+
+        if (fieldName.includes('change_pct')) {
```
drg101, line 272: maybe there should be a comment here that explicitly states what this is hard coding?

src/js/ui/widgets/Popup/PopupColorInfo.js
```diff
@@ -66,6 +66,7 @@ import * as d3 from '../../../third-party/d3.min.js';
 import { temporalId } from "../../../library/Constants";
 
 export default React.memo(function PopupColorInfo({ colorFieldName, colorSummary, obj }) {
+    console.log(colorSummary);
```
drg101, line 69: remove this

PR #444, opened 2022-01-28
Make preliminary MACA data available

pierce-smith1:
Makes MACA future climate predictions available as `Future Climate (Tract/County)`.

This is just 1 of 40 models that will eventually be ingested and there are some issues with the underlying dataset (two variables are missing and had to be left out of Aperture), so I will need to update this eventually, but everything that's _here_ should work fine.

Test all the variables of the new dataset, make sure coloring works, make sure it shows up across the entire country, and let me know if anything looks weird. (Remember that the tract version of the dataset is missing some tracts because the spatial resolution of the data isn't high enough)

### DIFF
--- a/src/js/library/Query.js
+++ b/src/js/library/Query.js
@@ -632,8 +632,6 @@ const Query = {
             senderID: id
         });
 
-        console.log(body);
-
         const responseListener = this._getRawDruidQueryListener(query, geometryData);
         this.queryWorker.addEventListener("message", responseListener);
     },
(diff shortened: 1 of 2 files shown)


Marc-9:
For both Future Tract/County no queries actually load
Looking at the network traffic we see the query was a success (we get response headers) but the body is always empty.

Marc-9:
Looks good, did county over the continental US and did not see any issues.
The data in the widget looks fine, but I believe Shrideep wants us to convert from kelvin to C or F I don't remember, perhaps this will be a front end issue

Marc-9 (approved):

PR #445, opened 2022-02-16
Metadata for general aqi

honeybadgerofdoom:
Colors/graphing for the general aqi score. The dataset in aperture is called `EPA AQI Score`.
Some thoughts on this
- Aperture by default uses a date range for this dataset. This is wonderful in the inspection pane when you want to look at the graph of aqi over time for a given county
- The colors on the map only relate to the _first_ date in the date range. To me, this is less than intuitive.

I do think this is ready to merge, unless anyone finds a bug/other issue. I'd love to hear other opinions on the map coloring. Because this dataset does have both a range and a single-date visual component, I'm not sure about the best way to address this.

### DIFF
--- a/src/js/library/Color.js
+++ b/src/js/library/Color.js
@@ -128,6 +128,16 @@ export default class Color {
             this.overrideColor = predefinedColor.colorCode;
             return;
         }
+        else if (predefinedColor.style === "aqi_special") {
+            const green = Array(10).fill("#00e400");
+            const yellow = Array(10).fill("#ffff00");
+            const orange = Array(10).fill("#ff7e00");
+            const red = Array(10).fill("#ff0000");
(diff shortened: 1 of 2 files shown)


drg101 (approved):

pierce-smith1 (commented):
> The colors on the map only relate to the first date in the date range

What do you mean by this? I'm able to select between average, first, last, min, max, etc for the AQI field, and the coloring updates accordingly. Does this mean the averages, maxes, mins, etc are calculated for just the first date? If so why?

Also: are large gaps in the available counties like this expected? 

image

honeybadgerofdoom:
@pierce-smith1 yes, those are counties without sensors. And again, yes - you can make those selections, I just hadn't noticed.

PR #446, opened 2022-02-22
Update total datasets display

honeybadgerofdoom:
Shrideep asked me to make this change to Aperture. It just removes the total_datasets field from the dataset explorer when the search bar is inactive. This is a 1-line change.


Marc-9 (approved):
this works

PR #453, opened 2022-04-06
Remove Heatwaves from menu

honeybadgerofdoom:
Shrideep asked me to remove this dataset, as it doesn't currently work. I saved the metadata for it in the LegacyCode folder.


drg101:
Why do we need an `_old` if the old stuff is in the commit history?

honeybadgerofdoom:
Because it's waaaaaaay easier to find

Marc-9 (commented):
Heatwave was my favorite dataset I will miss it
Also Im with Daniel Legacy folder is silly, why re-invent the git wheel

honeybadgerofdoom:
@drg101 @Marc-9 I hear ya, though Shrideep did explicitly ask for a LegacyCode folder...

drg101:
> @drg101 @Marc-9 I hear ya, though Shrideep did explicitly ask for a LegacyCode folder...

This sounds completely unmaintainable -- every time we make a change are we supposed to add it to some other folder?

honeybadgerofdoom:
No, just every time we delete something we might want later. For example, in the likely event that we add a Future Heatwave dataset in the future, we can just grab the metadata from the legacy folder. In the case of the Modeling code, it's there as a reference for anyone working on Delphi.

drg101:
> No, just every time we delete something we might want later. For example, in the likely event that we add a Future Heatwave dataset in the future, we can just grab the metadata from the legacy folder. In the case of the Modeling code, it's there as a reference for anyone working on Delphi.

@honeybadgerofdoom  This is what git releases are for. If someone wants to use the future heatwave dataset in the future they can browse the code of a release in which it existed.

PR #464, opened 2022-04-15
Lexicographic sort of advanced constraints

### LINKED ISSUE #460: Sort constraints in advanced tab

tmoleary21:
Overall, pretty simple additions. First pull request!

I added sorting in the `extractLayerConstraints` function in Layer.js because it is when the array of all of the layer constraints is first created. I originally tried sorting in the AdvancedConstraints component, but then there are issues with the index not matching between `allLayerConstraints`, `defaultLayerConstraints`, and `activeLayerConstraints`.

Closes #460


pierce-smith1 (commented):
Not sure this is working as intended? (This is the Public Schools dataset)

image

tmoleary21:
Uh oh, that is definitely not right. I'll mark this as a draft and see if I can figure out why it didn't work right.

tmoleary21:
All right that should be fixed. Thank you for noticing it @pierce-smith1. The datasets I was using to test it weren't very representative of the whole. It now sorts by the label property instead of just name

pierce-smith1 (approved):
LGTM

PR #467, opened 2022-04-21
Added coloring support for logarithmic datasets

### LINKED ISSUE #465: Add choropleth coloring support for logarithmic datasets such as county population

### LINKED ISSUE #466: Change default color gradient to go from Cool -> Hot colors, and make colors darker

victorberggren:
closes #465, closes #466

* Only the county_total_population dataset is defined as a logarithmic dataset in this PR, so use that set to test the changes. To make a different dataset use a log scale, add "style": "log_gradient" to its "color" field in menumetadata.json.

I made a new file called logarithmicMapping.js, which exports a single function that takes a value within a range [min1, max1], and maps it logarithmically onto a range [min2, max2]. This function is used when normalizing the value from the dataset to fit the color gradient array which has length 100, and is also used when calculating the positioning of the black bar inside the color gradient in the Inspection Pane.

Next step is to figure out which datasets need to be rendered with a log gradient as default, then add a toggle to the Inspection Pane to let the user switch between a linear and logarithmic color gradient for any dataset.


drg101 (approved):

drg101:
Looks good. Only thing I notice is the "reverseGradient" option doesn't work on this.

drg101 (commented):
Why the opacity change?

victorberggren:
> Why the opacity change?

With the original opacity the colors are very hard to see on the topography and satellite map layers.

drg101:
Did someone request that the opacity change?

victorberggren:
Yeah Shrideep requested the opacity change -- I demo'd it in last week's meeting

drg101:
Alright then I'll merge it

PR #468, opened 2022-04-26
Add search bar for advanced constraints menu

### LINKED ISSUE #462: Add Search bar to advanced tab

tmoleary21:
This adds a search bar to the layer's advanced constraints pop up menu. The constraints are filtered simply based on whether they contain the search string, ignoring case.

The components hold the indices of the constraints in a state. This is to ensure that filtering preserves the position in the overall array, so the checkbox click will select the correct constraint. So the index list is what is actually filtered.

Closes #462


tmoleary21:
Since this is still waiting, I've converted to a draft until #470 is merged. That way I can use the SearchBar.js component it adds to clean this up.

tmoleary21:
Ready for review. Now uses the generalized SearchBar component added in #470.

drg101 (approved):

PR #503, opened 2022-06-30
Standardizing Gradient

falarrieu:
Purpose of pull request is to standardize all gradients. Default gradient is to be blue for the lower bound, and red for the upper bound.


tmoleary21:
Why are we getting rid of reverse gradient? Just curious.

falarrieu:
Shrideep asked me to standardize all gradients to start in blue and to end in red. I would assume it would fall along the lines of users having the consistent understanding that blue is always the lower value.

honeybadgerofdoom:
Did Shrideep ask to specifically remove the ability to set a reverse gradient?

We could achieve this PR with only the changes in `menumetadata.json`, without removing the _ability_ to set a reverse gradient in the future. Seeing as `reverseGradient` is something that needs to be specified in `menumetadata.json`, the default behavior would be blue=low, red=high.  I don't know if we should axe this feature completely unless Shrideep asked us to.

This looks good to me - just want to sort that out before we proceed.

falarrieu:
It is now defaulted only by removing reverse gradient in menumetadata.json. What ideas to we have for it in the future?

honeybadgerofdoom (approved):
Looks great!

tmoleary21 (approved):

PR #505, opened 2022-07-04
Bug Reporting

honeybadgerofdoom:
Add UI Component
Add hashing for auth so GH doesn't nuke the PAT

Test this by clicking on the lil `bug` icon and submitting a bug. You should then see your submitted bug in the `issues` tab.


honeybadgerofdoom:
Dealing with conflicts in `package-lock.json`...

PR #507, opened 2022-07-04
Bug Report

honeybadgerofdoom:
Add UI Component
Add hashing for auth so GH doesn't nuke the PAT

Test this by clicking on the lil `bug` icon and submitting a bug. You should then see your submitted bug in the `issues` tab.


tmoleary21:
Looks great, it's nice and easy to use. When you send the issue, it appears on GitHub as submitted by your account. Is this ok? Is there any way to get it to be generic or a sustain account, or would that take creating a new GitHub account?

PR #509, opened 2022-07-05
Additions to DateSlider options

tmoleary21:
This builds off Matt's previous PR that introduced the option to use a date slider for time based data sets. I added a switch to turn on cumulative data, which essentially just fixes the minimum to equal the absolute minimum. So sliding just adds or removes data to the view instead of it being independent. I also added a way to choose a range of months when stepping by year. These new options are in an accordion labeled "Other options."

This also updates the state correctly to match the slider when other controls are changed. There is another bug currently with the slider that causes it to repeat the same year on two consecutive tick marks. I think this is caused by leap years, so when the step is added to get the next point, it isn't enough to jump the whole year.

I'd like to come back to this soon to create an object to hold most of the state data for all of the time components. I think this will improve readability and make it more expandable, and maybe make it easier to fix that bug.


Kmbear3:
Looks good to me! I think maybe Matt should approve it too just because this is piggybacking on his work.

honeybadgerofdoom (approved):
This works great!

PR #515, opened 2022-07-08
New layers should be enabled by default

tmoleary21:
This changes the workspace so when a new layer is added, the switch is on by default and is rendered on the map. It only auto enables when there are no other enabled layers.


tmoleary21:
Converted to draft. I forgot we wanted it to only auto enable if there are no other layers in the workspace already.

honeybadgerofdoom (changes requested):
Two things here

- If I add a dataset to the workspace, it turns on as it should. If I then remove the same dataset and add it again, it does not turn on by default. If I do this again, it does turn on, every other time.

- If I use the bulk-add feature (the '+' icon) and add multiple datasets at once, they all turn on automatically. This could be problematic - it's a lot of querying all at once. Maybe change this so that bulk-add doesn't turn on datasets by default?

tmoleary21:
Working on this. Having some issues fixing it for @honeybadgerofdoom's  requests, but I'll hopefully have it done today.

tmoleary21:
This should be ready for review @honeybadgerofdoom. Let me know if you have any more issues that I didn't see.

As far as I'm concerned, a lot of the changes in this pull request are temporary. The change needs to be put in place, but the best way to implement this will take more time. I think this will be much better implemented once we can refactor most of the workspace state into a custom hook. A big issue I had completing this was that the logic for the workspace is very spread out. So I think consolidating into a hook would make things much simpler

honeybadgerofdoom (approved):
looks great!

PR #518, opened 2022-07-08
Bug report includes active layers.

honeybadgerofdoom:
This PR just adds state to the Bug Report. It now includes which datasets were active when the report was submitted.


Kmbear3 (approved):
Dope!

PR #522, opened 2022-07-12
Bubble chart

falarrieu:
The main purpose of this pull request is to add the implementation of the bubble chart. While implementing I came across some difficulties in the original FrameControls.js. It was mostly hard-coded in a way that would only function for statistical charts using two variables. The bubble chart requires three, the third being for the bubble value. I was able to reduce some of the logical complexity as well as making it more dynamic/scalable with objects.


tmoleary21 (changes requested):
Wow! Nicely done, this looks great. Could we remove the console logs?
There are also a lot of CodeClimate issues. I'm sure we'll end up ignoring most of them, especially since you did this a lot like the other charting templates. Maybe we can take a look at improving readability when we're improving the rest of charting.

tmoleary21 (approved):

falarrieu:
Oops! Removed the console logs. I agree I think we could do something about the readability.

honeybadgerofdoom (approved):
Looks good to me!

PR #527, opened 2022-07-28
Update dynamic charts

falarrieu:
Implemented the ability to dynamically add and remove charts under specified chart type.


src/js/ui/charting/ChartingImprovements/NewFrameControls.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import Grid from "@material-ui/core/Grid";
+import ChartDropdown from "../ChartDropdown";
+
+
+export default function NewFrameControls(props) {
+    const {setConstraint, setConstraint2, setConstraint3, numDropDowns, options} = props;
+    const dropdownOptions = ["X-Axis",  "Y-Axis", "Bubble Variable"];
+    const objects = []
```
tmoleary21, line 9: "objects" doesn't really describe what this is used for to me. Does it only refer to the dropdowns or can it be any controls? Could we rename it to reflect this?

src/js/ui/charting/ChartingImprovements/NewFrameControls.js
```diff
@@ -0,0 +1,51 @@
...
+
+    function createObjects() {
+        for(let i = 0; i < numDropDowns; i++){
+            objects[i] = {
+                name: dropdownOptions[i],
+                constraint: settingObjectConstraint(i)
+            }
+        }
+    }
```
tmoleary21, line 18: Do these functions need to be nested, or could they be pulled out of the parent function? I try to avoid using nested functions because they make huge parent functions, and its hard for readability. Maybe they could be pulled out into a class/object? If you think they need to stay, leave them.

tmoleary21 (changes requested):
NewFrame and NewFrameControls should probably just be called Frame and FrameControls. They won't be new forever.

So I requested a few changes, but this looks nice. I'm sure we'll need to make changes in the future as we/you add more chart types. Right now I'm thinking another custom hook, similar to the useChartingTabs hook will be nice for the actual charts

tmoleary21 (approved):
Thanks for making those changes. I think we can talk about nested functions in a meeting later on. I don't feel that they're a huge deal but I try to avoid them. I'd like to hear what you think about it, so we can just talk about it next time we meet.

Looks good 💯

PR #528, opened 2022-08-02
Pane open and close button +refactor

tmoleary21:
The main purpose of this pull request is to create a button to open/close the pane so there is another way besides clicking a map feature. I implemented this with a button that follows the left side of the drawer as, kind of like a pull tab. The pane is now called the "Analysis Pane." When opened with the tab, it defaults to the charting tab, since no object should be selected at this point. If a map feature is clicked, it will automatically switch to the inspection tab.

This PR also includes some refactoring.
* Moved inspection pane specific functions, state, and useEffects to the inspection pane component from the analysis pane component
* Restructured files. Folder inside src/js/ui is now called dataAnalysis. This contains a Charting and Inspection folder for the appropriate files.
* The charting folder under src/js/ui is (mostly?) untouched. This will need to be removed later when we remove the old charting window.


falarrieu (changes requested):
Small UI problem occurs when selecting to add a new tab followed by closing the drawer.

falarrieu (changes requested):
There is now an issue occuring in typeTabs when clicking on a new part of the map while having some chart tabs added.

tmoleary21:
Just added a `PersistentTabPanel` component that uses the visibility CSS attribute to hide the tab contents. This preserves state and keeps tab contents rendered even when hidden. We need to discuss whether this is the right course of action. 

It's good that we can keep the state local to components and that it will preserve all state of the tab contents between tabs. I don't love that everything is rendered at all times, it feels wrong. I think it would be ok to go back to conditional rendering if we create hooks and objects to hold the charting states and place it in the `ChartingPane` component. It just wouldn't make sense to hold all the individual chart states in the higher component unless it was organized with some sort of data structure.

falarrieu (approved):
Looks good! A lot of problems fixed.

PR #529, opened 2022-08-03
Fix chart updates

falarrieu:
Small issue where charts were not re-rendering enough to keep up with data change. The fix included lifting the chartData state into RenderTabPanels.js.  Also just some small refactoring and cleanup.


tmoleary21 (approved):
:+1

PR #532, opened 2022-08-03
Migrate Histogram and Radar charts

tmoleary21:
I haven't moved the line graph over yet because it was only every used for COVID. It might be something we move over much later, if we ever even implement it for anything else.

Radar is just as broken as it has always been. That's a task for another PR.

I also removed COVID from tabs.


tmoleary21:
I added the collapse to the histogram real quick. It doesn't collapse the extra controls yet, but I think I'll get that working in my next PR. Radar chart will take some work to get that working

falarrieu (approved):
Looks good!

PR #533, opened 2022-08-10
Revamp boxplot

falarrieu:
Improvements made to older box plot chart. The changes include; fixed min/max/median horizontal lines, changed to vertical to remove text overlap of tick marks, and added label of variable type. There is one change the was made to Frame Controls which is to account for the case that if only one drop down is being created then render drop down with "Variable."


tmoleary21 (approved):
Looks great! I'm going to add one more task to trello for the boxplot. It gets a little scrunched with social vulnerability index since it ranges from 0 to 1. Icebox for now.

falarrieu:
That is something that might be fixed if we decide to do the zoom functionality.

PR #535, opened 2022-08-13
Graphing tract and county together

tmoleary21:
As the name says, this PR implements the ability to graph tract datasets with county datasets. This is applicable for scatter plots, bubble plots, and correlograms. I'm going to wait to finish correlogram until it is ready.

This also makes gridmet datasets graphable, kind of. It seems like it allows historical climate to be graphed but not future. It looks like future isn't even being sent through the chart system. Adding both historical tract and county to the workspace will break it as well, but I'm not sure why yet. Seems like something also in the chart system. Graphing them at all was made possible by switching the way we match data together into points (x,y,z). Gridmet data doesn't have a location name field.

Before, data was matched up based on the name of the location. Not only did this make graphing tracts and counties together impossible, it also means we actually miss a lot of data because location names aren't unique. Data would be replaced with whatever place was calculated last with the same name. Now, we match based on GISJOINs, which are unique. They also make it easy to match tract with county because you can get the GISJOIN for the county that contains the tract by truncating the tract GISJOIN to 8 characters. [Thanks Pierce](https://stackoverflow.com/c/project-sustain/questions/35).


falarrieu (approved):
Looks good!

PR #536, opened 2022-08-14
Close buttons for chart tabs

tmoleary21:
This adds an X button on each tab to close them. When a tab is closed, nothing is lost in the associated tab panel. When the same tab is reopened the same content will be there. I thought this would be the best way to do it because it preserves the user's work. So closing tabs becomes more about removing clutter.

There are more changes here than were originally planned because removing tabs caused a bug to arise with the MUI tabs that made the scroll buttons behave weirdly. I found it was because of the div we had that was displaying as a flex box. Learned from here: [https://github.com/mui/material-ui/issues/23711](https://github.com/mui/material-ui/issues/23711).


falarrieu (approved):
Looks awesome!

PR #537, opened 2022-08-14
Frame Controls Class

falarrieu:
This pull request is to implement the FrameControlsClass. This is mainly to reduce conditional rendering in the FrameControlsClass but still allowing for future changes to controls for their relative chart.


tmoleary21 (changes requested):
Love the idea of using some inheritance with this! Could we rename FrameControlsClass to just FrameControls?

tmoleary21 (approved):
Looks great!

PR #538, opened 2022-08-14
Correlogram revamped

falarrieu:
This pull request is mainly to implement the Correlogram chart. With a few additions such has; refactoring, disabling menu items, fixing formatting in box plot.


tmoleary21 (changes requested):
I like the idea of disabling variables that are already being used in the dropdowns, but it makes it very difficult to swap variables to a different axis. I think it's important this is possible. Can you change it so it only disables the one in the current dropdown? It would be even better if when you select a variable that's already in a different dropdown it would remove it from that dropdown. Maybe you could add a method for that in the FrameControls class

Otherwise nice! I definitely think we need to have the rows and columns labeled at least with the ABCD labels soon, but we can leave that for now.

tmoleary21 (approved):
👍

PR #542, opened 2022-08-17
Fix ungraphable datasets that should be graphable

tmoleary21:
This really just makes changes in graphPriority.json, because that is (almost) all that determines whether a collection shows the graphable icon in the workspace. Median age (county) is now graphable. The issue was that it was trying to collect data that didn't exist in the dataset. I changed the names to what do exist and it now works.

Things that don't work:
* Future climate. It says it is graphable in the workspace search, but there is no entry for it in graphPriority. I can't remove the symbol for now without reworking a lot of workspace code, so for now it stays wrong.
* Point datasets. These were weird and I will come back in the future to get them working. Basically, no matter what bounds are used, they all get filtered out so no data is sent. I haven't been able to figure out exactly why yet.
* There were a few entries in graphPriority for datasets that don't exist in our system. So I moved them to the broken file.


falarrieu (approved):
Looks good!

PR #543, opened 2022-08-18
Improved Charting UI

tmoleary21:
Freddy and I have finished the new charting UI! This pull request will be very large as we have been using this branch as our master for about a month now. Please take a look through the charts and send us any comments or concerns you may have. I wouldn't worry about looking through the code too much, we have been scrutinizing each other's code along the way. So just if you have the time. We would like to have this merged by 2pm tomorrow (Thursday 8/17). 

The new UI should be easy to find since the old button still works. Basically, everything has been moved to the right side drawer, previous home to the inspection pane. It is now called the analysis pane, and has tabs for the inspection pane and charting. Each tab within charting is a different type of chart, and multiple of each can be added. Many of the charts have been revamped to work better and the controls for all of them have had many improvements. 
One improvement is the ability to chart tract and county together on most chart types. The only one you can't is correlogram, since it needs same length arrays to calculate correlation.

A few datasets that wouldn't graph before are now graphable, such as the historical climate and county median age.

I would like to seriously commend Freddy for all this work. He did the vast majority of chart revamping and migration, along with many other things. Really great work!


tmoleary21:
There will be more pull requests coming from us for further charting improvements in the near future, such as a revamped radar chart. These will be things that we decided weren't worth including just to push back release.

honeybadgerofdoom (approved):
This looks so good! Awesome work you guys!

PR #545, opened 2022-08-25
Revamp radar chart

tmoleary21:
Revamp radar chart in looks and include some notification text to let the user know it only works with a dataset with at least 3 graphable variables.

I created a class to do all the d3 operations. I'm not sure yet if it's something I will do in the future. I decided to wrap the instantiation in a useMemo so it won't recreate the object every time, and that's the biggest reason I may not do this again. My thinking was that to keep the d3 code clean, I'd have to split it into tons of functions that pretty much all take very similar parameters, so why not put those into an object that then handles it all. I think the code is cleaner, but we may be able to achieve the same effect by keeping the functions outside an object and just passing the common parameters into each as one object themself.

Future improvements:
- Text wrap
- Hovering tooltip for data polygons
- Allow graphing variables from different datasets at the same time


falarrieu (approved):
Looks good!

PR #548, opened 2022-08-30
Beginning workspace refactor

tmoleary21:
Here is the beginning of the workspace refactor, which is the beginning of the aperture refactor. We are not planning to have a separate dev branch for the refactor, instead opting for incremental changes in each pull request that will not break anything.

In this PR, I have created two hooks: `useWorkspace` and `useLayers`, which is called within `useWorkspace`. They both instantiate a class that holds state that previously existed in the Workspace component. The state is held as members of the instantiated object. The idea here is to have all actions performed by the workspace exist as functions in these classes. So the only time the values of the state variables should be used outside the class should be if it needs to be displayed on the screen, or maybe if it's used to display something else.

This PR is to create a starting point so Freddy and I can start adding in these functions.

Let me know what you think of this, and how you'd feel coding with it! Whole idea is to make it more readable and expandable.


falarrieu (approved):
Looks good! Let's see what we can do.

honeybadgerofdoom (approved):
Looks great to me! Love the custom hooks.

PR #575, opened 2022-10-06
Flarrieu data flow workspace

falarrieu:
This PR just sets up the static methods in LayerManager. Now state from Workspace such as adding/removing collection and enabling/disabling layers is communicated to the LayerManager.


tmoleary21 (approved):
👍

PR #583, opened 2022-10-08
Flarrieu add/remove collections

falarrieu:
Changed the logic behind adding and removing layers to when there are enabled and disabled rather than being added to or removed from workspace.


tmoleary21 (approved):
👍

PR #585, opened 2022-10-09
Flarrieu add remove collection to from deck layer

falarrieu:
Once a layer from the workspace layers is enabled then we will add the corresponding collection to appropriate deck layer. If a layer is disabled remove it from specified deck layer.

To meet these requirements we must handle the case of when the first layer is added to the workspace it will be already auto-enabled.


tmoleary21:
Looks good! Nice progress. 
I'm getting this "error," which I think is coming from deck gl
image
Any idea why? It doesn't seem super urgent to fix though.
Can we keep the console logs for now? They're kind of nice to have right now I think.

tmoleary21 (approved):
👍

PR #598, opened 2022-10-19
Clickable Charts

absarah:
- Potentially implement a rerender to the top of the Analysis Panel when one of the lower charts are clicked (eg: Correlogram)
     - Avoid having to scroll to the top


src/js/ui/dataAnalysis/Charting/ChartingPane.js
```diff
@@ -62,7 +62,7 @@ import RenderTabPanels from "./RenderTabPanels";
 import TypeTabs from "./TypeTabs";
 
 export default function ChartingPane(){
-    const chartTypes = ["Histogram","BubblePlot","ScatterPlot","Box Plot","Correlogram","Radar"];
+    const chartTypes = ["Histogram","BubblePlot","ScatterPlot","BoxPlot","Correlogram","Radar"];
```
tmoleary21, line 65: This change to the BoxPlot name is good for consistency, but it causes the box plot to not render correctly. I think it should just be making the same change in the switch statement in Frame.js. Could you take a look at that?

tmoleary21 (changes requested):
Works great! You're probably right that having it move the scroll back to the top would be ideal, but if you'd rather us look into that down the line, I don't see any issue with that. Once the box plot thing is fixed, ready to merge

src/js/ui/dataAnalysis/Charting/ChartingPane.js
absarah: Yes absolutely, I'll jump into Frame.js and correct that render issue. For consistency reasons, I'll work through that file to make the names consistent across the program.

absarah: I will also work on the rerender ability with a useEffect() to avoid extra work in the future. =)

falarrieu (approved):

PR #601, opened 2022-11-11
Kass data subsetting

Kmbear3:
This adds the ability for the clients to change the available datasets for visualizations.


Kmbear3:
Tis broken, closing

PR #602, opened 2022-11-11
URL parameterization for curating datasets

Kmbear3:
This PR allows the user to subset the data that is available for display in the workspace. 

- To test, append: /?curatedSet=<name of array located in curatedLayers.js>

This should then only display the collections that are listed in the array that is specified in the array.


tmoleary21 (approved):
Works really great! This is a cool feature.
It would be cool if users could save curated sets to their account kind of like they can save workspaces, or maybe when they save workspace to specify whether the datasets they selected should be the only ones showing on load. Something interesting to discuss in the future maybe.

PR #621, opened 2022-12-16
Histogram chart option in inspection pane

tmoleary21:
This implements a switch that can change the chart type in the inspection pane from line to bar. The bar chart x axis is shown as dates. To make it a histogram, the dates really represent the start of the time interval of the bar.

Improvements for later:
- The tooltip isn't working for some reason I haven't figured out
- The time ranges on the x axis could be improved visually

I decided to stop spending time trying to work on these because we need to get this out.
Originally, this was Abby and I working together. Thanks to her for the hard work!


falarrieu (approved):
Awesome work y'all!

